PR #96, opened 2025-08-11
Use artifacts

### LINKED ISSUE #95: Tether force should be plotted
To be able to properly compare the simple model with the ram model, the tether force has to be plotted.

1-Bart-1:
Uses artifacts.toml in precompilation and closes #95


1-Bart-1:
Replaced by #103

PR #148, opened 2026-02-24
Add helper scripts

ufechner7:
- [x] add run_julia
- [x] add install
- [x] add Manifest-v1.12.toml.default
- [x] remove empty Manifest-v1.10.toml.default
- [x] update Manifest-v1.11.toml.default
- [x] add workspaces to the main Project file
- [x] cleanup the test/Project.toml file and add the main project as source
- [x] add the main project as source to examples/Project.toml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ AGENTS.md
 CLAUDE.md
 Manifest.toml
 !/Manifest-v*.toml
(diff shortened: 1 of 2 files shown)


Project.toml
1-Bart-1: Julia 1.10 is not supported

PR #197, opened 2026-04-21
Update vsm call and update vsm

1-Bart-1:
And make sys_struct.set mutable, such that it can be updated, for instance if the system structure is deserialized but the settings were created new.


ufechner7 (approved):

1-Bart-1:
Already fixed in main

PR #216, opened 2026-05-31
Fix init tether len

1-Bart-1:


ufechner7 (changes requested):
apply_tether_init_stretched_lens! currently treats only STATIC points and winch points as non-translatable boundaries. Because tether_downstream_idxs can traverse through other tethers' segments, a different tether’s start/anchor point can still be translated if it is DYNAMIC and not a winch point, contradicting the docstring intent (“other tethers' anchors stay put regardless of their point type”). Consider adding all tether start_point_idx values to boundary so anchors never get dragged when another tether is scaled.

ufechner7 (approved):

PR #217, opened 2026-06-01
Fix multi tether 2

1-Bart-1:
## Summary

Redesigns tether initialisation so a tether is positioned by an explicit **stretched** standoff and pre-tensioned by a target **force**, with the unstretched rest length derived. Also fixes multi-tether clustering for tethers that share a rigid wing.

Stacked on #214 (`fix-makie-panning`).

## New model

Per-tether inputs (both optional):

- `init_stretched_length` — the stretched standoff, i.e. the placed point geometry (Σ‖segment‖). Default = the geometric/CAD length (no-op when unset). Drives placement.
- `init_tether_force` — target initial spring force [N], default `0`.

`reinit!`:
1. scales positions so each placed tether sits at its `init_stretched_length`,
2. derives the unstretched rest length from the force law (zero-velocity, tension branch): `len = stretched · (1 − force / unit_stiffness)` (force `0` ⇒ `len = stretched`, zero tension).

So you specify *where the kite sits* (stretched) and *how hard the line pulls* (force); the rope rest length follows. `init_unstretched_length` is no longer an input — it is derived — and specifying it in YAML is an error.

## Root detection (anchor-agnostic)

A tether is *placed* only if one endpoint is a ground-fixed boundary (`STATIC` point or a winch point); that endpoint is the fixed anchor — **either `start` or `end`**, so the usual `kite → ground` definition works. Scaling runs from the anchor toward the free end, rigidly translating everything downstream of the free end. A tether with neither endpoint anchored is an error.

Tethers that share a `RIGID_DYNAMICS` wing are recognised as one structure via **rigid-body connectivity** (not just segment connectivity — rigid wings have no inter-point segments) and placed as a single cluster. When several roots feed one structure, the cluster is placed by the **mean displacement** of all roots (averaging length *and* direction), emitting an `@info` (gated on `prn`).

`init!`/`reinit!` gained an `apply_tether_lengths` kwarg to skip placement.

## Breaking changes

- `init_unstretched_length` (YAML) removed → error; use `init_stretched_length` + `init_tether_force`.
- `Tether` gained `init_tether_force`; `init_stretched_len`/`init_unstretched_len` are now `Union{SimFloat,Nothing}` (`init_unstretched_len` derived); the positional length constructor arg (now the stretched length) is optional. Serialized models must be rebuilt.

Version bumped to **0.11.0**; see `CHANGELOG.md`.

## Tests

`test/test_tether_init.jl` covers single/multi-root placement, the anchor-agnostic root rule, the non-root error, the loop-to-anchor error, and force-derived rest length.


ufechner7 (approved):
What is a standoff?

PR #223, opened 2026-06-08
Improved psys check

1-Bart-1:


1-Bart-1:
bigger pr incoming, that includes this fix

PR #232, opened 2026-06-16
Fix warnings

ufechner7:
Fix warnings about unused parameters

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ docs/src/assets/2plate_kite_structure.png
 docs/src/assets/tether_sys_struct.png
 docs/src/assets/pulley_sys_struct.png 
 mwes/mwe_01.jl
+data/catenary_export.yaml
--- a/src/yaml_loader.jl
+++ b/src/yaml_loader.jl
@@ -310,15 +310,15 @@ function parse_aero_mode(text::String)
 end
 
 """
-    load_wing(mode::AbstractAeroModel, row, idx, data, set, wing_type, vsm_set,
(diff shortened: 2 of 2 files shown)


1-Bart-1:
Instead of renaming the unused parameters to `_data` / `_twist_surfaces`, consider **de-naming the slots** — drop the binding entirely and keep only a type annotation:

```julia
function load_wing(mode::AbstractAeroModel, row, idx, ::Any, set, wing_type,
                   vsm_set, yaml_to_ref, yaml_parse_ref_points,
                   yaml_parse_origin, ::Any)
```

These two positional slots only exist so this method's signature matches the `AeroPlate` method (which does use them), so there's nothing meaningful to name. `::Any` keeps the dispatch slot, removes the binding, and the linter has nothing to flag — no `_` prefix and no need for the explanatory docstring paragraph.

For the mapping closures that ignore their argument (`row -> set`, `row -> vsm_set`, `row -> wing_type`, `row -> mode`), the closure must still accept the row the caller passes, so a slot is required there. The equivalent there is the bare placeholder rather than an `_`-prefixed name:

```julia
:set     => _ -> set,
:vsm_set => _ -> vsm_set,
```

`_` alone is Julia's canonical throwaway (not a prefix on a real name), and StaticLint doesn't flag it.

If the `_`-prefix noise is bothersome across the whole file, the other option is to disable just this lint rule via `"julia.lint.useoffuncargs": false` — these are false positives (args required for dispatch / calling convention), not real unused values, so suppressing the rule is defensible and needs zero source changes.

1-Bart-1:
outdated

PR #258, opened 2026-08-01
Networkdynamics backend

1-Bart-1:


1-Bart-1:
won't work, replaced by #261

PR #264, opened 2026-08-11
Make pulley damping a per-pulley variable

ufechner7:
The damping of the pulley velocity was a hardcoded MTK parameter (`@parameters pulley_damp = 5.0`) with no way to reach it. It is now a `damping` field of `Pulley`, read in the equation as `params.pulleys[i].damping`, so it goes through the flat-parameter sync and can be changed on a running model without a rebuild.

Settable with `set_pulley_damping(sys, damping[, pulley_idxs])`, as a `damping` keyword of the `Pulley` constructor, or as an optional `damping` column of the YAML `pulleys` table. Default unchanged (5.0).

The added field changes the serialized layout of `Pulley`, so the patch version is bumped: caches keyed on the package version miss cleanly instead of failing to deserialize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # CHANGELOG
 
+## Unreleased
+
+### Added
+- Per-pulley `damping` [1/s]: the damping of the pulley velocity is a field of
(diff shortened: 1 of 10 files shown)


1-Bart-1:
fixed in #261

PR #266, opened 2026-08-12
Save and restore the full state through a SysLog

1-Bart-1:
Branches off `aeropressure-per-section-inflow`, so it also carries 8f41d820 (per-section inflow for `AeroPressure`), which is not mine.

## What this adds

A `SysLog` row now holds everything the integrator needs to restart, so a settled or mid-flight state can be saved and resumed instead of re-solved.

`update_sys_state!` writes point velocities, per-frame body turn rates, twist_surface twist and rate, and pulley length and rate alongside the positions and orientations it already wrote. `update_from_sysstate!` reads all of them back, rebuilding the body rates into the **principal frame** the ODE actually integrates (bodies are logged in the body frame). It no longer zeroes point velocities or caps twist surfaces at four.

Because that is the same function `replay` uses, replays now show real velocities and per-surface twist instead of zeros.

Restart with `init!(sam; remake=false, reinit_sys=false)` — `reinit_sys=false` keeps `init!` from calling `reinit!(sys_struct, set)`, which would reset positions from the CAD frame and discard the loaded state.

Requires KiteUtils ≥ 0.11.14 (OpenSourceAWE/KiteUtils.jl#127), which carries the new columns and the `precision` parameter.

## What this removes

`save_sys_struct_to_yaml`, added earlier on this same branch in 72bd8ed4. Writing a whole *structure* back to YAML had to guess at the CAD/world frame boundary: positions were written as `pos_cad`, because that is the frame aero sections are matched in and the frame `reinit!` derives `pos_w` from — which left a settled configuration needing a manual mapping, and a rigid wing placed by a `transforms` block never reloaded at all. Saving *state* has no such boundary. Net effect is ~40 lines of source in place of a 497-line writer and its 180-line test.

## Testing

`test_init!` round-trips every model the suite builds through a `SysLog` on disk and asserts `integrator.u` is unchanged, scrambling the structure via the `diff_vars` setter in between so a restore that did nothing cannot pass.

That shape matters: whether a component is part of the state at all depends on its `DynamicsType`, so checking `integrator.u` after a reinit proves the coverage without needing to enumerate which fields are differential.

Note on tolerance: the check is `rtol=1e-10`, not exact, because bodies are logged in the body frame and rebuilt into the principal frame, which costs a few ULP in the quaternion/matrix conversions.


1-Bart-1:
Superseded by #268, which carries this work — the `SysLog` full-state round trip and the loader columns it needs — after it was merged into that branch. Closing so the content lands once. The branch is left in place.

Issue #263, opened 2026-08-11
Slack segments are overdamped: compression_frac softens stiffness but not damping


1-Bart-1:
A slack segment keeps its full damping while its stiffness is softened to
`compression_frac`, so its damping-per-stiffness ratio is inflated by
`1 / compression_frac`. At the common `compression_frac = 0.01` a slack line is
100× overdamped relative to its stiffness — effectively a pure dashpot with
almost no spring. Physically a rope that carries no spring force should resist
no closing speed either, and the material's damping-per-stiffness ratio should
hold taut or slack.

`segment_spring_force` (`src/components/kernels.jl`), shared by both backends
via `segment_eqs!` (`src/generate_system/segment_eqs.jl`):

```julia
damping = unit_damping / len
stiffness = ifelse(len > l0, unit_stiffness / len,
                   compression_frac * unit_stiffness / len)
return stiffness * (len - l0) - damping * spring_vel
```

|            | stiffness      | damping   | damping/stiffness |
|------------|----------------|-----------|-------------------|
| taut       | `k/len`        | `c/len`   | `c/k` (material)  |
| compressed | `cf·k/len`     | `c/len`   | `(c/k)/cf`        |

### Why damping is not simply folded into `slack`

Moving `unit_damping` inside the `slack` factor was tried and reverted. It makes
the force **discontinuous** at `len == l0`: the stiffness term crosses
continuously only because `len - l0` vanishes there, whereas the damping term
jumps by `(1 - compression_frac) · unit_damping · spring_vel / len`. A taut
tether sits on that crossing, so the solver chatters across it.

Measured on `test_pulley` (5 mm Dyneema, `unit_damping/len ≈ 1900 N·s/m`): both
pulley testsets died with `AssertionError: Solver unstable at t=0.42: MaxIters`,
on the monolith and kernel backends alike. `test_tether_winch` drifted out of
tolerance the same way (45.20 vs 45.0, 55.27 vs 55.0). Reverting to the table
above made all of them pass.

### Suggested fix

Make the factor continuous and apply it to both terms, so
`damping/stiffness = c/k` on either side and the force has no jump:

```julia
slack = compression_frac +
        (1 - compression_frac) * clamp((len - l0) / band, 0, 1)
```

`band` needs a sensible default and probably wants to scale with `l0`. Inside
the band a loaded line sees reduced damping, but a tether under load sits well
[...]

PR #267, opened 2026-08-12
Control a segment's damping and a tether's compression behaviour

1-Bart-1:
Two related knobs for slack lines, both live in the right-hand side.

### `Segment(...; compression_damping_frac)`

Scales the damping term of a segment the way `compression_frac` scales the
stiffness:

```julia
taut     = len > l0
damping  = ifelse(taut, unit_damping, compression_damping_frac * unit_damping) / len
stiffness = ifelse(taut, unit_stiffness / len, compression_frac * unit_stiffness / len)
spring   = stiffness * (len - l0) - damping * spring_vel
```

The default `1.0` is the previous behaviour: damping is unaffected by
compression, so the force stays continuous at `len == l0`, but the damping
*ratio* jumps by `1/sqrt(compression_frac)` the moment a segment goes slack —
on a taut bridle that sits right at the crossing, so the effective damping
ratio flickers between two values.

Setting `compression_damping_frac = compression_frac` gives one damping ratio on
both branches instead. `compression_damping_frac = compression_frac = 0` makes a
slack segment carry no force at all, which is what a rope actually does, at the
cost of a force step of `unit_damping / l0 · spring_vel` at the crossing. The
trade-off is documented on `segment_spring_force` so neither choice is silent.

Addresses OpenSourceAWE/SymbolicAWEModels.jl#263.

### `Tether(...; compression_frac, compression_damping_frac)`

A Route 2 tether generated its segments with the `Segment` defaults, so an
auto-expanded line was stuck at `compression_frac = 0.1` and could not use the
above at all. Both are now forwarded to every generated segment, and both are
readable as `tethers` YAML columns.

Together with a winch-less tether holding its length fixed, this makes Route 2
the way to split one line into several segments without writing every segment
and intermediate point out by hand — which is how V3Kite now discretises its
bridle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@
   no force at all, at the cost of a force step of
   `unit_damping / l0 · spring_vel` at the crossing. Also readable as a
   `segments` YAML column.
+- `Tether(...; compression_frac, compression_damping_frac)` hands both to every
+  segment a Route 2 tether generates, so an auto-generated line is no longer
+  stuck at the `Segment` default of `compression_frac = 0.1`. Also readable as
+  `tethers` YAML columns. A tether without a winch holds its length fixed, so
+  this is the way to split one line into several segments without writing every
+  segment and intermediate point out by hand.
 - A `SysLog` now carries the whole differential state, so a simulation can be
   restarted from any logged step. `update_sys_state!` writes point velocities,
   per-frame body turn rates, twist_surface twist and rate, and pulley length and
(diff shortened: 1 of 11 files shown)


1-Bart-1:
Superseded by #268, which carries this work — `compression_frac`/`compression_damping_frac` on `Segment` and `Tether` plus the YAML unset-cell rule — after it was merged into that branch. Closing so the content lands once. The branch is left in place.

PR #271, opened 2026-08-18
Give the continuous aero modes the flow curvature moment

1-Bart-1:
A section rotating about its own spanwise axis sees a curved flow, and
thin-airfoil theory pays for it with the moment increment
`Δcm = -(π/4)·q̂`, `q̂ = q·chord/(2·v_rel)`. VortexStepMethod applies this
inside `solve!`, so `AeroDirect` and `AeroLinearized` already had it while
the continuous modes, which rebuild the panel force symbolically, did not.
Turning on the solver's `flow_curvature` flag therefore gave two different
models depending on the aero mode.

Off by default, and read from each wing's own `VSMEngine`, so no mode can
carry the term while another does not.

### Where the rate comes from

Not plumbed in from the solver: each panel derives its own rate from its
sections' trailing minus leading edge apparent wind. A deforming wing
therefore gets the true per-section rate, twist and flapping included,
instead of a projection of one rigid body rate.

### AeroPressure

`AeroPressure` spreads each panel's load over surface contour nodes and had
no couple to apply, so it dropped `panel_couple`. It now takes the increment
as a chordwise force pattern over those same nodes: thin-airfoil's `A₂`
loading, `sin 2θ`, offset so the weights sum to zero and their first moment
about the leading edge is `-1`.

`A₂` is the one chordwise mode carrying a moment and no lift, which is what
the increment must be — the lift of the curved flow already arrives through
the 3/4-chord collocation inflow, and the `A₁` parabolic-camber mode that
*generates* the effect is symmetric about mid-chord, so forcing its net
force to zero cancels its moment too. A contour whose nodes carry no first
moment now errors rather than producing garbage weights.

### Tests

`test/test_flow_curvature.jl` builds each mode with the flag on and off and
asserts the physics, not the output: zero rate and zero moment at rest,
blindness to uniform translation, agreement with `ω·y_airf` under rigid
rotation, and a moment opposing the rotation feeding it.

The non-rigid witness uses a spanwise rate changing sign *twice*
(`cos 2πs`). An antisymmetric left/right rate is not valid — `y_airf` itself
flips sign at mid-span, so a uniform rigid `ω` reproduces it exactly.

| | monolith | kernel |
|---|---|---|
| result | 60 passed | 34 passed, 8 skipped, 0 errors |
| `dM·ω` ContinuousAero | −66.4 / −69.1 W | — |
| `dM·ω` AeroPressure | −96.6 / −98.4 W | −97.1 / −98.9 W |
| rigid rotation vs `ω·y_airf` | 5.2% / 0.8% | skipped |
| sign-alternating twist, rigid residual | 0.999 / 1.0 | skipped |

The kernel backend has no observed-variable readout, so introspection cases
are skipped there and only the end-to-end moment is checked.

### Notes

- Requires `VortexStepMethod` 4.1.0 for the `flow_curvature` solver flag.
- `flow_curvature` enters `get_sys_struct_hash`; it changes the emitted
  equations, so a cached model built with the other setting must not be
  reused. Existing VSM-wing caches rebuild once.
- Branched off `fix-kinematic-body-restore`, which this targets; merge that
  first.


1-Bart-1:
Folded into #285, which now targets `main` directly and carries these eight commits. Closing so the flow curvature, unsteady aero and live polar work reviews as one PR.

PR #273, opened 2026-08-21
Read a fitted wing's pose back from its weighted reference points

1-Bart-1:
Follow-up to #270.

`KinematicWingReadout` collapsed each `WeightedRefPoints` to its first point, so a
`KINEMATIC` wing whose `origin_idx`, `z_ref_points` or `y_ref_points` name several
points reported the wrong `pos_w`, `vel_w` and `R_b_to_w` on the `KernelBackend` —
and the apparent wind and reported scalars derived from them. The dynamics already
used the weights; only the struct read-back did not.

The readout now carries the `WeightedRefPoints` themselves, and
`wing_kinematics_from_points!` blends through `get_ref_position_from_points`, the
same helper the monolith uses.

`test_yaml_weighted_ref.jl` gains a read-back check after a step: 16/16 on both
backends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@
   scene-mutating `plot!` stays on `Makie.plot!`.
 
 ### Fixed
+- `KernelBackend` reads a fitted (`KINEMATIC`) wing's pose back from the weighted
+  blend of its reference points, as `MonolithBackend` does. The read-back kept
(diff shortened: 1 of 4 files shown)


1-Bart-1:
Folded into #276: its commit is cherry-picked there unchanged. Both touch the `kernel_backend/state.jl` read-back and would have conflicted, so they review better together.

PR #274, opened 2026-08-21
Test the per-panel aero equations directly instead of reading the model

1-Bart-1:
The continuous-mode testset checked the per-panel reconstruction by reading
variables out of a compiled model, which only worked on `MonolithBackend` —
`ProbWithAttributes.sys` is `nothing` on `KernelBackend` — and keyed on names that
are codegen artefacts.

It now evaluates `panel_force_eqs` directly on the panel's own geometry and inflow,
substituting its equations forward in the order they are emitted. That checks the
equations the model compiles, on both backends, without touching a compiled model.
Polar coefficients are handled in two passes: symbolic placeholders first, then the
real `cl`/`cd`/`cm` at the resulting angle of attack.

Every intermediate is compared against VSM, not just the span sum, so a per-panel
error that cancels spanwise can no longer hide.

Measured worst-case error over all panels:

| | ContinuousAero | AeroPressure |
|---|---|---|
| chord, width | 0.0 | 0.0 |
| span axis | 1.1e-16 | 5.7e-17 |
| dynamic pressure | 3.8e-16 | 2.2e-16 |
| chord axis | 1.98e-4 | 0.0 |
| angle of attack | 1.9e-4 rad | 5.6e-17 |
| force | 1.0e-3 | 4.0e-16 |

The remaining `ContinuousAero` error is one systematic effect, not noise. VSM takes
the chord direction between two section blends weighted by *panel spacing* (`ncp`),
where `panel_force_eqs` uses the midpoint. The test measures that gap as `blend` and
bounds every axis-dependent tolerance by it — `blend` came out bit-identical to the
chord-axis error, which is what identifies it as the sole cause. On a uniformly
panelled wing `blend` is zero and the whole reconstruction is exact, which is why
`AeroPressure` is at machine precision.

The old `alpha_atol_deg=1.0` sat ~90× above that bias and could not have caught it.

A follow-up PR moves the spacing weight into `panel_force_eqs` and drops the `blend`
allowance.

86/86 on `MonolithBackend` and 86/86 on `KernelBackend`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@
   scene-mutating `plot!` stays on `Makie.plot!`.
 
 ### Fixed
+- The live aero modes (`ContinuousAero`, `AeroPressure`) take a panel's chord
+  direction between the two bounding sections blended by panel spacing, as
+  `VortexStepMethod` does, instead of at their midpoint. On a non-uniformly
+  panelled wing that midpoint tilted the whole airfoil frame: the angle of attack
+  was off by 1.9e-4 rad and the per-panel force by 0.1%, a systematic bias that
+  partly cancels spanwise and so never showed up in the wing total. Every
+  per-panel quantity now matches VSM to machine precision. The weights are
+  refreshed with the frozen circulation, so they follow the deforming mesh. They
+  are a new parameter, so a model cached before this needs `remake=true` once.
 - A tether with no winch takes its rest length from `params.tethers[i].len`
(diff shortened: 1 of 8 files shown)


1-Bart-1:
Superseded by #276, which now carries these two commits rebased onto `main` alongside the cold-start and `sync_params!` work. Combining them because they touch the same aero refresh chain (`refresh_particle_aero!`, `solve_and_freeze_circulation!`) and conflicted otherwise.

PR #278, opened 2026-08-22
Split the test suite over parallel CI jobs

1-Bart-1:
Each runner configuration now runs the suite as **three parallel jobs** instead of
one: 15 test jobs where there were 5.

- `test/suite_groups.jl` — the file inventory `runtests.jl` and the workflow share
  (excluded files, kernel skips, measured per-file runtimes), plus the split into
  groups of roughly equal runtime.
- `test/ci_matrix.jl` — crosses those groups with the runner configurations and
  prints the workflow matrix. Runs on a bare Julia, no dependencies.
- `CI.yml` — a `test-matrix` job computes the matrix; each test job gets its files
  as `test_args`.

Groups come out even: **724 / 724 / 723 s** monolith, **600 / 599 / 599 s** kernel.

## Why three groups and not one job per file

Measured from run 32564565156, ubuntu/1.11, no coverage:

```
09:19:04  buildpkg
09:20:37  Pkg.test starts
09:34:22    ✓ SymbolicAWEModels            683 s   ← precompile, src changed
09:43:35    ✓ SymbolicAWEModelsMakieExt    551 s
09:43:59  --> first test file              ← 23 min in
10:12:58  done                             ← 36 min of tests
```

**~23 min of every job is fixed** — precompiling the package, which the restored
depot cache cannot serve once `src/` changes. Average per-file test time is 49 s,
and the account's plan caps standard runners at **20 concurrent jobs**, so wall
clock ≈ total runner-minutes ÷ 20.

| Split | Jobs | Runner-min | Wall clock |
|---|---|---|---|
| today | 5 | ~280 | 74 min |
| **3 groups × 5 configs** | **15** | **~570** | **~38 min** |
| 1 job per file × 5 configs | 220 | ~5300 | ~4.5 h |

One job per file is a 4× *slowdown* here: each extra job buys 49 s of parallelism
for 23 min of runner time. Three groups is the point where every job still fits in
a single concurrency wave alongside `docs` and `reuse-lint`.

The machinery is per-file capable — `SYMAWE_TEST_GROUPS=file` in the workflow's
`env:` block gives one job per file, any integer gives that many groups — so the
granularity is a one-line change if the numbers move.

## Notes

- The depot cache key now names the Julia version and arch instead of the whole
  matrix row (`include-matrix: false`), so a configuration's three jobs share one
  cache entry instead of each saving its own ~2 GB copy of the same depot.
- Per-file runtimes were measured with the compiled-model `.bin` cache shared
  across the whole suite. Split into groups, files that rode another file's model
  build pay for their own, so `FILE_DURATIONS` is worth re-measuring from the
  first run on this branch.
- `Pkg.test(; test_args=[...])` now errors on a name that matches no test file,
  instead of silently running nothing — CI depends on those names being right.
- No cross-file dependencies in the suite: each `save_log`/`load_log` pair is
  self-contained, and nothing includes another test file.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,50 +29,37 @@ jobs:
       - name: Run REUSE lint
         run: docker run --rm --volume ${{ github.workspace }}:/data fsfe/reuse lint
 
+  # The matrix is one entry per runner configuration and test group, so that the
+  # runner configurations and the split live in `test/ci_matrix.jl` next to the
+  # suite instead of being duplicated here.
+  test-matrix:
+    name: Compute the test matrix
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.compute.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
+
+      - name: Compute the test matrix
+        id: compute
+        run: |
+          MATRIX=$(julia test/ci_matrix.jl)
+          echo "Matrix: $MATRIX"
+          echo "matrix=$MATRIX" >> "$GITHUB_OUTPUT"
+
   test-with-code-coverage:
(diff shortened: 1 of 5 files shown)


1-Bart-1:
Doesn't help much.

PR #283, opened 2026-08-24
Fix the panel moments: flow curvature, and anchor what the Cp pattern places

1-Bart-1:
Two halves of the same problem — the panel pitching moment was wrong when the wing
twists, and wrong in steady state too.

## Twisting: the flow-curvature increment

The continuous VSM modes re-derive the panel force symbolically instead of reading
`solve!`'s output, so VortexStepMethod's `flow_curvature` moment — the thin-airfoil
`Δcm = -(π/4)·q̂` of a section rotating about its own spanwise axis — was absent from
them while `AeroDirect` and `AeroLinearized` had it. Enabling the solver flag silently
meant two different models.

Each panel's rate now comes from its own sections' trailing minus leading edge
apparent wind, so a deforming wing gets the true per-section rate — twist and flapping
included — rather than a projection of one body rate. Off by default, and read from the
wing's solver, so no mode can have the term while another does not.

This half also brings apparent mass and a two-state Wagner lag on the particle VSM
modes, both off by default.

## Steady state: nothing anchored the moment

`AeroPressure` has anchored its **force** since it was written — each panel's node
forces get an equal share of `(panel_force - traction_net)`, so they sum to the live
VSM total whatever the Cp pattern does. The **moment** had no equivalent. The mode
assumed the frozen surface traction already carried the polar `Cm`, which holds only as
far as the Cp is trustworthy, and a pattern that misplaces load chordwise fed the
structure a wrong panel moment with nothing to notice.

Measured on an SK100 mid-span section at 5°, moment about the quarter chord:

| Cp source | delivered | polar `Cm` | error |
|---|---|---|---|
| NeuralFoil tables (what the project ships) | −0.0411 | −0.0655 | **37% short** |
| XFoil tables, same contour | −0.0711 | −0.0718 | 1% |

`couple_shape` is normalised to a chordwise first moment of `-1`, so node forces whose
first moment is `Σ x·f` already carry a couple of `-Σ x·f`. `freeze_traction_pattern!`
now stores the traction's first moment alongside its net, and `anchored_couple` credits
both that and the equal-per-node residual against `panel_couple`. The delivered moment
is then the VSM one for any Cp — the same guarantee the force already had.

The `AeroPressure` `scatter_couple` override is gone: it took only the flow-curvature
increment, on exactly the assumption this fixes. Both backends now scatter the whole
`panel_couple` and credit the traction explicitly.

## Backends

Only the wiring differs. The monolith adds the term symbolically; the kernel folds the
live half into the scatter's force weight (`share + shape·chord_mean`) and the frozen
half into the point offset, so no new kernel plumbing.

## Not verified locally

Testing is left to CI. The changed files parse and the struct's three constructor call
sites were checked against its 19 fields by hand; nothing else was exercised here.


1-Bort-1:
Superseded by #285. Every commit here now has a home in `main` or in `feat/live-polars`, so this closes rather than rebases — the branch has been CONFLICTING since 24 Aug and nothing unique is left on it.

The first four are the base of #285's stack (`2beae003`, `921f6f60`, `82f5b1b9`, `c602ba77`; two are patch-identical to the ones here, two differ only by rebase). The two that were unique:

**`cf04c3cc` "Trace VortexStepMethod's panel kernel instead of restating it"** landed on its own as #284, merged into `main` today as `95b62495`; `feat/live-polars` took it in when it merged `main`. In `src/aero_modes/common.jl` there, `panel_force_eqs` builds from `VortexStepMethod.panel_axes`, `panel_inflow`, `panel_force_directions`, `panel_loads`, `effective_alpha`, `dynamic_pressure` and `panel_couple_force`; the hand-written axes, the `atan` angle of attack, the lift and drag crosses and the dynamic-pressure line are gone, `panel_span_signs` and `store_chord_weights!` call `panel_span_vector` and `panel_chord_weight`, `smooth_norm` forwards to VSM's, and the `section_pitch_rate` / `flow_curvature_cm` twins are gone with their `private_functions.md` entries. The only difference is where the pitch rate enters: an extended `panel_inflow` call here, a local `pitch_rate_rhs` calling the same `VortexStepMethod.section_pitch_rate` there.

**`887a3787` "Anchor the AeroPressure panel moment the way its force is anchored"** is covered by `3890b2f6` "Place the panel's pitching moment, not only its force" in #285. Same defect and the same guarantee, by a stronger mechanism: `freeze_traction_pattern!` stores `traction_moment` — the frozen pattern's full moment about the panel's leading-edge midpoint — alongside `residual_arm`, and `pressure_couple` (`src/aero_modes/pressure.jl:868`) credits both against the target `ContinuousAero` would place, force at the quarter chord plus `panel_couple`, and places the deficit. Where this branch spread the residual on a scalar `chord_mean` over an equal `1/n` share, that one uses a share-weighted vector arm over `node_residual_share`, and divides by the airfoil frame's own triple product rather than assuming the axes are orthogonal on a billowed panel. It is also asserted, which this branch never was — `test/test_pressure_aero.jl`, "each panel's force and pitching moment are VSM's", to `rtol=1e-10`.

The `scatter_couple` override this commit deleted is gone on both sides; #285 re-signatured the hook to `scatter_couple(mode, slots, i, panel)` and routes `AeroPressure` through `pressure_couple`.

Porting the two forward would conflict in every hunk — the struct field block and its constructor, the `aero_component` call site (`build_panel_force_eqs` returns six values now and `surface_node_forces` takes `node_residual_share`), the `freeze_traction_pattern!` accumulator, `accumulate_point_offset!` and `aero_scatter_entries` — and would land the same physics on the weaker arm. Nothing to carry over.

PR #284, opened 2026-08-24
Trace VortexStepMethod's panel aerodynamics instead of restating it

1-Bart-1:
## Why

`panel_force_eqs` held a hand-written symbolic copy of VortexStepMethod's per-panel force — the airfoil axes, the chord blend, the angle of attack, the lift and drag directions, the force and the pitching-moment couple. Correct, but it matched the numeric solver only as long as someone kept both sides in step by hand. VSM's recent billow-axis fix is the kind of edit that changes one side and not the other.

## What

VortexStepMethod.jl#265 pulls that algebra into `src/panel_aerodynamics.jl`: pure, branch-free, number-type-generic functions of the section geometry and the flow. This PR builds the equations by *calling* them with symbolic arguments:

```julia
axes   = VortexStepMethod.panel_axes(le_1, te_1, le_2, te_2, chord_weight, orient)
inflow = VortexStepMethod.panel_inflow(axes, va_1, va_2, vind)
dirs   = VortexStepMethod.panel_force_directions(axes, alpha[i], spanwise)
loads  = VortexStepMethod.panel_loads(bound_axes, bound_dirs, q_dyn[i], cl, cd, cm, scale)
```

The physics is no longer written here. What is left is the part that is genuinely ours — **where to tear the expression graph**. Each stage binds to its slot variable before feeding the next, so the emitted equations keep the same structure as before: `alpha` still reads the raw axes, `dir_lift`/`dir_drag` still read the slot `alpha`, `panel_force` still reads the slot directions and widths. The chord blend keeps its offset form (the one that avoids the 3.7× symbolic build cost) — that reasoning moved into the shared docstring, where it now applies to both consumers.

`smooth_norm`, `panel_span_signs` and `store_chord_weights!` go through the same functions.

## Backends

Both backends already routed through `panel_force_eqs` — the monolith via `build_panel_force_eqs`, the kernel backend via the `AeroPanel` component — so they pick this up together. `Backend read-back parity` passes.

## Tests

`test_aero_modes`, `test_backend_parity`, `test_continuous_aero`, `test_flap_aero`, `test_pressure_aero`: **932 passed, 0 failed, 0 errored.**

## Symbolic build cost

The concern with rewriting these expressions is MTK build time, so it was measured rather than assumed — `test_aero_modes`, same machine, same session state, seven models:

| model | main | this branch |
|---|---|---|
| `aero_modes_1` | 0.796 s | 0.700 s |
| `aero_modes_2` | 0.377 s | 0.399 s |
| `aero_modes_3` | 0.486 s | 0.469 s |
| `aero_modes_4` | 0.548 s | 0.473 s |
| `aero_modes_5` | 0.821 s | 0.926 s |
| `aero_modes_6` | 0.779 s | 0.842 s |
| `aero_modes_7` | 0.511 s | 0.536 s |
| **total `structural_simplify`** | **4.32 s** | **4.35 s** (+0.6 %) |
| **total model init** | **62.6 s** | **62.0 s** (−0.9 %) |

Within run-to-run noise, which is what the preserved tearing is there to achieve.


1-Bart-1:
Rebased onto current `main` (was conflicting since 2026-08-25) and verified.

`src/aero_modes/common.jl` merged cleanly; the only conflict was `CHANGELOG.md`,
where `main` had gained the per-point-wind entries. Both `### Changed` bullets are
kept. `main`'s `panel_force_eqs` signature is unchanged from what this branch was
written against, so nothing was dropped in the rebase.

### Verification

`test_aero_modes.jl` is the right test for this change: its strict driver compares
the model's total aero force and moment against a full `VortexStepMethod.solve!` on
the realised panel geometry, which is exactly the contract that de-duplicating the
kernel has to preserve.

**779/779 pass**, across all 7 (aero mode × dynamics type) combinations, run against
the **registry** VortexStepMethod 4.3.0 rather than a patched checkout — so this is
a clean behaviour-preservation result, independent of any in-flight VSM change. The
parity margins are comfortable, e.g.

```
[direct particle]      rel_F=0.0026   (tol 0.006, 43.3% of budget)
[direct rigid]         rel_F=0.000139 (tol 0.001, 13.9% of budget)
[continuous particle]  rel_F=0.000923 (tol 0.006, 15.4% of budget)
```

### Why this matters beyond the de-duplication

VortexStepMethod.jl#272 found that `panel_axes` built the panel normal from the
leading-edge step while the bound vortex, `width` and `y_airf` all came from the
quarter-chord step. `panel_force_eqs` on `main` carries a hand-written copy of that
same expression (`z_cross = x_unit × (le_1 - le_2)`), so the fix would have to be
applied twice and kept in step by hand — and a third copy exists in
`wagner_reference_frame` on the branches that add it.

With this merged, VortexStepMethod.jl#273 propagates here for free, and the class
of drift it fixes cannot recur.

1-Bart-1:
@1-Bort-1 fix the test failure.

1-Bort-1:
Picked up as `SymbolicAWEModels.jl-284`, on the branch this pr already has.

1-Bort-1:
### Fixed

CI never reached a test: `UndefVarError: SMOOTH_FLOOR not defined in
VortexStepMethod`, while precompiling. The panel kernel this PR traces first
ships in VortexStepMethod **4.2.0**, so the compat bound here is right — but
both tracked default manifests still pinned **4.1.1**, and CI seeds its manifest
from those (`Pkg.instantiate` only warns on a stale `project_hash`; it never
re-resolves). `Manifest-v1.12.toml.default` is workspace-wide, so the pin
reached `test/` as well; the 1.11 one is root-only, which is why that job
resolved 4.3.1 for itself and was the single green test job.

### Changes

- `Manifest-v1.1{1,2}.toml.default` — `VortexStepMethod` 4.1.1 → 4.3.0, version
  and tree hash only: the dependency list is byte-identical between the two, so
  nothing else moves. Regenerating them instead takes fifty other packages
  forward, which is what put the monolith jobs over the 180-minute cap on
  `feat/live-polars` (`1be6c744`). 4.3.0 rather than the 4.2.0 minimum because
  that is the version this PR's parity numbers were measured against.
- Rebased onto `main` — v0.16.0 was released after this branch was cut, so the
  diff was reverting `version` to 0.15.2. The entry now sits under a fresh
  `## Unreleased` instead of inside the released section.
- `src/generate_system/helpers.jl:81` — `smooth_norm`'s docstring had grown a
  why; it says what it does now (CLEAN_CODE §4).

### Verification

- [x] On the box against the pinned 4.3.0: `test_aero_modes` 779,
      `test_backend_parity` 3, `test_continuous_aero` 52, `test_flap_aero` 18,
      `test_pressure_aero` 60+20 — **932 pass, 0 fail, 0 error** · GitHub CI:
      running · local mirror: started
- [ ] `deno fmt` not run — `deno` is not on this box; no added line exceeds 92
      chars, and REUSE lint passes on GitHub. No new or renamed public symbol,
      so no docs page changes.
- Risk: the pinned set is now ahead of what the defaults' `project_hash`
  describes, so `Pkg.instantiate` warns. It does not re-resolve, but a
  `Pkg.resolve()` would, and that pulls the fifty packages in with it.

### Commits

- `4b588484` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `39cfad60` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `ab99949d` Put the entry under Unreleased and say what smooth_norm is

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/284/files/ffd817600e8b1c5a399255c4589a9367d659b0bd..ab99949dc3579b461dc72fee1d381c03b5741649)

Task `SymbolicAWEModels.jl-284`

src/aero_modes/common.jl
```diff
@@ -343,45 +350,35 @@ function panel_force_eqs(slots, i, sections, flow, polars, spanwise, scale,
...
+        x_airf[:, i] ~ axes.x_airf;
+        y_airf[:, i] ~ axes.y_airf;
+        z_airf[:, i] ~ axes.z_airf;
+        v_eff[:, i] ~ inflow.v_eff;
+        alpha[i] ~ inflow.alpha;
+        q_dyn[i] ~ VortexStepMethod.dynamic_pressure(rho_1, rho_2, inflow.v_span);
+        dir_lift[:, i] ~ dirs.dir_lift;
+        dir_drag[:, i] ~ dirs.dir_drag;
+        panel_force[:, i] ~ loads.force;
+        panel_couple[:, i] ~ VortexStepMethod.panel_couple_force(
```
1-Bort-1, line 379: MINOR: The pitching moment is assembled by a separate panel_couple_force call while the PR says panel_loads already returns 'the force and the pitching-moment couple'; either loads carries a never-used moment or the same quantity now has two assembling codepaths.

Project.toml
```diff
@@ -98,6 +98,6 @@ TOML = "1"
 Tar = "1.10.0"
 Timers = "0.1.5"
 UnPack = "1.0.2"
-VortexStepMethod = "4.1.1"
+VortexStepMethod = "4.2"
```
1-Bort-1, line 101: MINOR: The compat floor is "4.2" but the added calls (panel_axes, panel_chord_weight, panel_couple_force, SMOOTH_FLOOR) must exist in 4.2.0; with the Manifest on 4.3.0 a resolver could still select a 4.2.x that lacks them.

1-Bort-1 (commented):

**Good**

- Eliminates the hand-synced symbolic copy of VSM's per-panel aerodynamics so the numeric solver and the emitted equations share one source (satisfies §2).
- Build cost is measured rather than assumed and the tearing claim is borne out: structural_simplify +0.6% and init −0.9%, within noise.
- Docstrings are rewritten to 'what' with the chord-blend offset rationale moved out of code, matching §3/§4.
- panel_span_signs, store_chord_weights!, and smooth_norm are all unified onto the VSM kernel, not just panel_force_eqs.
- The changelog entry states the behavior change and dependency bump compactly.

**Not good**

- `src/aero_modes/common.jl:379` — The pitching moment is assembled by a separate panel_couple_force call while the PR says panel_loads already returns 'the force and the pitching-moment couple'; either loads carries a never-used moment or the same quantity now has two assembling codepaths.
- `Project.toml:101` — The compat floor is "4.2" but the added calls (panel_axes, panel_chord_weight, panel_couple_force, SMOOTH_FLOOR) must exist in 4.2.0; with the Manifest on 4.3.0 a resolver could still select a 4.2.x that lacks them.
- Project.toml compat is "4.2" while the Manifest lands on 4.3.0 and the newly-traced functions must exist at the 4.2.0 floor; if they landed in 4.3 the lower bound is wrong.
- smooth_norm's default eps is flipped from the literal 1e-12 to VSM's SMOOTH_FLOOR for a general-purpose helper, silently touching every default-using caller if the constants differ.
- store_chord_weights! drops the n_panels < 2 early return, so a single-panel wing now depends on panel_chord_weight(nothing, w, nothing) returning 0.5.
- bound_axes is rebuilt by collecting slot columns, then panel_couple_force re-reads raw chord[i]/width[i]/q_dyn[i] instead of the bound fields it was given.

1-Bort-1:
**Local full suite: PASS** (25 min, Julia 1.12.7, one cell of the matrix)

PR #285, opened 2026-08-25
AeroPressure: polars regenerated from the deformed shape, not a flap angle

1-Bart-1:
Requires VortexStepMethod 4.3, which is released; compat and the default manifests are on it.

This was stacked on #271 (`flow-curvature`); it is now one PR against `main` and carries those commits too.

## Live polars

The SK100's chord is a Timoshenko beam, so its trailing edge bends into a smooth camber shape. Reading the polars on a flap angle δ asks that shape to be one number — and the number depends on which node pair you measure between. On the SK100's settled state the chord ends read 0.6–8.4°, while the pair straddling the 0.75c hinge reads −0.5–2.8° for the same deformation. Neither is wrong; the question is ill-posed.

`AeroPressure(; live_polars=true)` stops asking it. Each VSM solve:

1. The structural points a panel already scatters its load onto are read as **control points**. Their offset off the deformed chord line, minus the reference offset recorded at build, is the camber increment.
2. That deforms the panel's Kulfan fit analytically — a matvec against a constant CST basis, never a refit.
3. NeuralFoil is evaluated a few angles either side of the panel's own α, in one batched pass over the whole wing.
4. Those values are written straight into the panel's own polar table.

There is no fit and no new aero model: a live polar is an ordinary `POLAR_VECTORS` table whose knots and coefficients are rewritten in place. Between two samples the polar is the straight line between two exact points, so there is no fit residual to bound. δ then carries no information, so it is dropped from the generated equations entirely and only α stays live in the RHS.

The chord's own rotation and stretch need no handling: `update_vsm_wing_from_structure!` already took the panel frame from the deformed LE/TE, so what is left is exactly the residual camber the CST basis can represent. `chord_frame_coordinates` is the one place that frame is written down, shared with `freeze_traction_pattern!`, so a control point and the traction pattern cannot disagree about which way is up.

The traction pattern is regenerated from the same deformed shapes once the solve converges — the contour offset by the camber increment, `Cp` from one batched pass at the converged α, skin friction from the flat-plate closure at each panel's Reynolds. Previously a deformation changed how hard a panel pulled but not where.

### Limits, stated up front

- **The wing must carry chordwise stations.** With only LE/TE points the chord frame is built on the control points themselves and the deflection is identically zero; the build `@warn`s rather than pretending.
- A live polar spans only the angles it was sampled over and holds its last value past them, so bring-up wants damping enough to keep the solve inside that range — on the SK100, `start_world_damping` 300 rather than 20. Past the range the answer is stale rather than wrong-signed, and `polar_drift` warns.
- Post-stall equilibria are out of reach by construction: a local expansion cannot follow the branch jump.
- `check_live_polar(mode, wing; panel_idx)` runs one XFoil solve against the polar a panel is actually flying, defaulting to the panel NeuralFoil is least sure about. A confidence says the network is extrapolating; this says whether the extrapolation is any good.

Settled at 12 m/s the live and tabulated polars agree. At 20 m/s the live polars settle to a peak tube bending margin of 1.02 against the tabulated 4.71, with 3 of 49 joints past the collapse knee rather than 10.

## Flow curvature and unsteady aero (from #271)

The continuous VSM modes carry VSM's `flow_curvature` moment increment, the thin-airfoil `Δcm = -(π/4)·q̂`. They re-derive the panel force symbolically instead of reading `solve!`'s output, so the term was absent from them while `AeroDirect` and `AeroLinearized` had it — enabling the solver flag silently meant two different models. Off by default and read from the wing's solver, so no mode can have the term while another does not.

Unsteady aerodynamics on the particle VSM modes, off by default: `apparent_mass` gives nodes the air they entrain (thin-plate `ρ·π·c²/4` per unit span, in a new field rather than `extra_mass` — entrained air resists acceleration but has no weight), and `wagner` adds the two-state lift lag.

## Two load-path fixes

**A panel's load is shared between the two stations it lies between**, rather than rounded onto the nearer one. Rounding is a discontinuity, and two panels mirroring each other across the wing land on spanwise places agreeing to within a few ulps but not exactly — so the nearer station could be a different one for each, moving a whole panel's load a station across the span. A symmetric wing at a symmetric pose carried 1.14 N·m of yaw and 4.88 N·m of roll that nothing cancelled; both are now identically zero, with the pitching moment intact. A panel straddling a station also took the mean of two weights measured against different station pairs, which means nothing averaged.

**`AeroPressure` places each panel's pitching moment, not only its force.** The scatter anchored the force to the VSM panel force and left the moment to whatever the frozen `Cp` pattern integrated to, so the polar's `Cm` never reached the structure. Both the force and the moment now come out exact rather than approximate.

## BREAKING: a twist surface is a station

In the code and in the structural geometry YAML — `twist_surfaces:` and the per-wing `twist_surfaces: [...]` list are both `stations:`, `TwistSurface` is `Station`, and the file is `station_eqs.jl`. The entity is the structural points lying in one chordwise plane; twist is one of the things it carries, alongside the spanwise stations the aero load and the live polars are read on. It had grown three names — `twist_surface`, `station`, `strut` — which is how the load path and the deformation path came to compute the same blend twice without anyone noticing they were the same quantity.

## Tests

`test/test_live_polar.jl` (147 assertions): the build fits a base airfoil per panel and records the control-point frame, no panel is wired to a flap surface, a refresh leaves every panel on a table sampled about its own α, the chord-frame maths agrees with a hand-computed frame, a chordwise deformation moves the polar while the undeformed shape puts it back, the installed values are the network's own answers, and the traction pattern follows the deformed shape. Plus a check against the tabulated mode and against XFoil.

All four aero modes now fly the same patched fixture, so the yaw cases compare like with like.


1-Bart-1:
Note on CI: running this branch against a dev'd VortexStepMethod 4.2.0 locally, `test_aero_modes.jl` shows one failure — `pressure particle` yaw antisymmetry, `abs(moment_pos[3]) > 1e-3 * norm(force_pos)` evaluating 0.196 > 0.227. Everything this commit adds is gated behind `live_polars`; with it `false` the only observable change is `aero_hash_id` gaining a `false`, which is a cache key. So the failure is either already on `flow-curvature`, or it comes from the 4.1.1 → 4.2.0 bump this branch's compat forces — the combination this repo has not run before. Worth resolving before merge; I could not isolate it locally (two precompiles of the same SymAWE version from different trees deadlock on the same cache slug).

1-Bart-1:
Exercised on the SK100 (44 refined panels, chordwise Timoshenko beam), and it found the cold start: `lr.alpha_dist` is all zeros before the first solve, so every panel's polar was fitted about α = 0 while the wing flew at 5–11°, three fit windows away, and the solve failed. `live_polar_alpha` now seeds from each panel's geometric angle when there is no previous solve to read (`b15110e`), and upstream a `TAYLOR` polar is continued linearly past its window rather than following the polynomial's arms.

With that, on `system_ecm.yaml` at 12 m/s:

- 9–13 control points per panel, at chord fractions 0.116 … 0.996 — the chord beam and the pressure receivers give it real chordwise resolution, so the "no chordwise stations" warning does not fire.
- solve converges, α 4.93–10.84°, drift 0.46 of the fit window.
- against the same geometry's tabulated `(α, δ)` polars:

| | live | tabulated |
|---|---|---|
| Cl range over panels | 0.726 – 1.288 | 0.768 – 1.337 |
| Cm range over panels | −0.068 – 0.042 | −0.054 – 0.067 |

about 5 % apart, which is the Kulfan refit's non-uniqueness rather than a difference in physics. The durable fix is for the offline generator to write the weight vector the tables were made with, so the live mode starts from the same one instead of re-deriving it from the written contour.

Cost, warm, 44 panels: 0.166 s per refresh — that includes the drift guard's refit-and-re-solve cycles, not just the 4 ms polar regeneration. Worth measuring in a run before trusting it at a 0.1 s cadence.

`test/test_live_polar.jl` still passes 28/28 with the cold-start change.

1-Bart-1:
**Correcting the 0.166 s figure above — it was JIT, and the explanation with it was wrong.** Warm, on the same 44-panel SK100 wing, best of five:

| | tabulated | live |
|---|---|---|
| `refresh_particle_aero!` | 6.42 ms / 12.07 MB | **11.44 ms** / 18.46 MB |
| ├ `freeze_traction_pattern!` | 5.53 ms / 12.0 MB | same |
| ├ polar refit (NeuralFoil) | — | 3.59 ms / 6.22 MB |
| ├ deflection extraction | — | 0.085 ms / 159 kB |
| └ one VSM solve | 0.23 ms / 0 B | 0.26 ms / 0 B |
| drift-guard rounds | — | **1** |

Live polars add about 5 ms, essentially all of it the network's forward pass. The drift guard takes one round in steady state and a solve is 0.26 ms and allocation-free, so "the drift guard's refit-and-re-solve cycles" was simply not what the time was. Against `dt: 0.1` this is 11 % of the budget, not slower than real time.

`b886716` came out of checking it: `chord_frame_coordinates` was copying the leading edge points and both frame axes into fresh heap vectors on every call, ~1450 times a refresh. Reading them off the stack took deflection extraction from 0.142 ms / 390 kB to 0.085 ms / 159 kB. Tests still 28/28.

For scale on the remaining allocation: the 6.22 MB is `nn_forward` building a fresh matrix per layer, twice over for the symmetry flip. It is upstream of this and would need an in-place network eval to remove — worth doing only if the refresh ever reaches the critical path, which it has not: `freeze_traction_pattern!` in the existing mode allocates twice as much.

1-Bart-1:
@1-Bort-1 investigate why the tests are failing and fix if failure is found.

1-Bort-1:
Picked up as `SymbolicAWEModels.jl-285`, on the branch this pr already has.

1-Bort-1:
From #291, which is stacked on this branch: `docs/make.jl` fails here, not only under the stack.

`private_functions.html` renders at 506.77 KiB against the `size_threshold = 500 * 1024` in `docs/make.jl`, so `makedocs` throws `HTMLSizeThresholdError` and the build exits 1. The ~57 lines of new private symbols on this branch are what took it over: the two entries #291 adds are 2.0 KiB of the page (measured in the built HTML), and the overflow is 6.77 KiB.

`exported_types.md` (141 KiB) and `private_kernel_backend.md` (264 KiB) are still only past the warn limit.

1-Bort-1:
### Fixed

`test_flow_curvature.jl` errored on Ubuntu (both Julia versions) because
`rigid_rate_residual` fits the rigid rate through the normal equations, and their
Gram matrix is rank deficient whenever the panels share a spanwise direction —
which the flat `AeroPressure` fixture's do, so `gram \ moments` threw
`SingularException`. That throw skipped the line that takes the injected twist
field back off `sam_on`, so the next testset compared a poisoned state against a
clean one and reported the increment as wildly asymmetric in the rate. One cause,
two red tests.

### Changes

- `test/test_flow_curvature.jl:226` — the least-squares `omega` comes from
  `pinv` of the design matrix, which gives the minimum-norm fit when the axes
  span fewer than three directions instead of throwing. The residual it reports
  was always well defined; only `omega` was not unique.
- `test/test_flow_curvature.jl:233` — regression testset on four identical axes.
  The old body throws `SingularException(1)` on it; I ran that to check.

### Verification

- [x] `test/test_flow_curvature.jl` green on the box: 76/76 (CI had 74 pass, 1
      fail, 1 error). `AeroPressure` damping is now −99.3 W against −101.0 W for
      ω_y = ±3 rad/s, where CI read −54.2 against −263.6.
- [x] GitHub CI green on `1be184c4`, all seven checks, both previously red jobs
      among them (run 33996966620).
- The local mirror (`agent ci-local`) was started but the box reaped it under
  memory pressure, so it has no result to report.
- Risk: none — the change is a test helper's linear algebra. No model, no
  equation and no tolerance moved.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..1be184c40bbe9d305830db1310e5ff6aa493f3b7)

Task `SymbolicAWEModels.jl-285`

.github/workflows/CI.yml
```diff
@@ -32,7 +32,9 @@ jobs:
   test-with-code-coverage:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ matrix.backend || 'monolith' }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 120
+    # Temporarily wide so a monolith job runs to completion and prints its
+    # per-model build times; drop back to 180 once the slow build is found.
+    timeout-minutes: 1000
```
1-Bort-1: MAJOR: A temporary 1000-minute timeout plus a "drop back later" comment is a debugging leftover shipped in the PR; a failing job now hangs CI for ~17 hours, and the flag will rot without ever being reverted.

1-Bort-1 (commented):
<sub>Superseded by the review further down this thread.</sub>

CHANGELOG.md
```diff
@@ -2,7 +2,94 @@
 
 ## Unreleased
 
+### Changed
```
1-Bort-1: MAJOR: The entire file is mechanically re-wrapped to 92 chars in the same diff as the feature, so hundreds of reflow lines bury the actual additions; §1 requires formatting to be its own cleanup PR so the diff is skimmable.

CHANGELOG.md
```diff
@@ -2,7 +2,94 @@
...
+  `twist_surfaces: [...]` list are both `stations:`. The entity is the
+  structural points lying in one chordwise plane, and twist is one of the things
+  it carries, alongside the spanwise stations the aero load and the live polars
+  are read on. It had grown three names — `twist_surface`, `station` and `strut`
+  — which is how the load path and the deformation path came to compute the same
+  blend twice without anyone noticing they were the same quantity.
+  `TwistSurface` is `Station`, `station_control_points` says what it returns,
+  and the file is `station_eqs.jl`.
+
+### Changed
```
1-Bort-1: MINOR: A second consecutive ``### Changed`` header under the one ``## Unreleased`` block — the section accumulated duplicate headers instead of one consolidated block, exactly the accretion §7 warns about.

1-Bort-1:
**Local full suite: PASS** (25 min, Julia 1.12.7, one cell of the matrix)

.github/workflows/CI.yml
1-Bart-1: @1-Bort-1 revert this

CHANGELOG.md
1-Bart-1: @1-Bort-1 fix this

1-Bort-1:
### Fixed

The two threads you replied to: the temporary CI timeout is reverted, and the
`## Unreleased` changelog block now has one section per kind.

### Changes

- `.github/workflows/CI.yml:35` — plain `git revert` of `cb904f2e`, so
  `timeout-minutes` is 180 again with the comment it had. The longest job
  (kernel) took 119 min on the last run, so that leaves 61 min of headroom.
- `CHANGELOG.md` — `## Unreleased` had **three** `### Changed` blocks with
  `Fixed` and `Added` interleaved. Now one of each, in the order §5 and the last
  two releases use. A pure move: the 143 non-header content lines are
  byte-identical before and after.

### Verification

- [x] GitHub CI green on `1be184c4`, all seven checks (run 33996966620) · local
      full suite PASS there too. Neither new commit touches code, so I did not
      re-run the suite; CI re-runs anyway and is what exercises the 180-min cap.
- Risk: the restored comment says the kernel backend needs "~100 min". It took
  119. Still inside 180, but drifting — I left it as the revert put it rather
  than quietly editing a revert.
- **Correcting my docs comment above**: the Documentation job is green on this
  branch, so `docs/make.jl` does not throw here. That 506.77 KiB was measured
  under #291's stack; the subtraction I did to claim it overflows on this branch
  alone was an inference, not a build. `agent retract` would not take the
  comment back, so it stands corrected rather than removed.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..dac326f264fef1a748eb1a57d5a941705aea3e51)

Task `SymbolicAWEModels.jl-285`

CHANGELOG.md
```diff
@@ -16,70 +17,221 @@
...
+
 ## v0.15.2 22-08-2026
 
 ### Added
-- `reinit!(sam, integrator; solver=integrator.alg, kwargs...)`, which defaults the
-  solver to the one the integrator was built with. The three-argument method takes
-  it as a required argument, and a caller passing a bare `FBDF()` made the monolith
-  backend compile its whole right-hand side again at `ForwardDiff.Dual`.
+
+- `reinit!(sam, integrator; solver=integrator.alg, kwargs...)`, which defaults
```
1-Bort-1: MAJOR: The entire changelog from v0.15.2 down was re-wrapped to a narrower margin with no semantic change, burying the ~150 lines of real Unreleased entries under ~700 lines of reflow; §1 requires mechanical formatting to ship in its own cleanup PR so the diff stays skimmable, and this defeats that.

CHANGELOG.md
```diff
@@ -16,70 +17,221 @@
...
+  `twist_surfaces: [...]` list are both `stations:`. The entity is the
+  structural points lying in one chordwise plane, and twist is one of the things
+  it carries, alongside the spanwise stations the aero load and the live polars
+  are read on. It had grown three names — `twist_surface`, `station` and `strut`
+  — which is how the load path and the deformation path came to compute the same
+  blend twice without anyone noticing they were the same quantity.
+  `TwistSurface` is `Station`, `station_control_points` says what it returns,
+  and the file is `station_eqs.jl`.
+
+- TEMPORARY: `SymbolicUtils` is taken from a fork
```
1-Bort-1: MINOR: The 'TEMPORARY: SymbolicUtils is taken from a fork' note records a `[sources]` entry the rubric flags as something never clean and Bart's call; a workaround and its tracking issue belong in the PR description, not in the changelog a user reads for release deltas.

.github/workflows/CI.yml
```diff
@@ -32,7 +32,8 @@ jobs:
   test-with-code-coverage:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ matrix.backend || 'monolith' }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 120
+    # The kernel backend needs ~100 min of the budget and grows with the suite.
```
1-Bort-1: MINOR: The inline comment explains why the timeout grew (kernel backend budget) rather than describing the setting; per §3/§5 that rationale belongs in the PR, and a config comment tied to a specific value will read as residue once the budget changes again.

1-Bort-1 (commented):
<sub>Superseded by the review further down this thread.</sub>

CHANGELOG.md
1-Bart-1: The symbolicutils change is merged so no need to use a fork anymore

CHANGELOG.md
1-Bart-1: fix this

1-Bart-1:
@1-Bort-1 fix conflicts and address comments.

1-Bort-1:
### Fixed

The conflict with `main`, which had traced VortexStepMethod's panel kernel
through the same equations this branch was adding flow curvature to.

### Changes

- `src/aero_modes/common.jl` — `panel_force_eqs` keeps main's tracing and this
  branch's terms: the pitch rate, the curvature increment, the Wagner
  `alpha_eff` and both couples now come from `VortexStepMethod`'s
  `section_pitch_rate`, `flow_curvature_cm`, `effective_alpha` and
  `panel_couple_force`. The local twins of the first two are deleted, with their
  two entries in `docs/src/private_functions.md`.
- `CHANGELOG.md` — rebuilt from main's file, so `## v0.16.0` is what main
  released and `git diff origin/main -- CHANGELOG.md` is **145 insertions, 0
  deletions**: the reflow you flagged is out of the diff.
- `Project.toml` — kept `VortexStepMethod = "4.3"` over main's `"4.2"`. The
  `SymbolicUtils` `[sources]` entry stays; see the reply on that line.

### Verification

- [x] 1282 assertions green on the merge: flow curvature 76, aero modes 779,
      live polars 147, continuous 52 + 92, pressure 80 + 20, backend parity 3,
      per-point wind 33. The curvature increment is unmoved — `dM·omega` −99.3 W
      and −101.0 W, `rel_error` 1.59e-16 against VSM. · GitHub CI: running
- [ ] Docs build, REUSE lint, Julia 1.11: not run — `deno`, `reuse` and Julia
      1.11 are not installed on this box. CI is what covers them.
- Risk: `panel_force_directions` is fed the panel's bound axes rather than the
  unbound ones main passes, which keeps `b7a1ee31`. Same value, smaller
  expression graph, and the one place the resolution differs from main by
  choice rather than by necessity.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..f698c9f11eb40d001b9f8eb9d183351318d24c60)

Task `SymbolicAWEModels.jl-285`

1-Bort-1:
**CI: `Julia 1.12 - macOS-latest - aarch64 - monolith - pull_request` failed** on `f698c9f11`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34131083073/job/101771097538

Project.toml
```diff
@@ -56,6 +56,10 @@ MakieControlPlots = "6d616b69-6563-4f6e-8472-6f6c706c6f74"
 [extensions]
 SymbolicAWEModelsMakieExt = ["GeometryBasics", "MakieControlPlots"]
 
+# Until the fix in JuliaSymbolics/SymbolicUtils.jl#1052 reaches a release.
+[sources]
+SymbolicUtils = {url = "https://github.com/1-Bart-1/SymbolicUtils.jl", rev = "fix/isequal-dag-memo-v4.45"}
```
1-Bort-1: MAJOR: The [sources] entry pinning SymbolicUtils to a personal fork is explicitly never clean per §9 (Bart manages how local packages resolve), appears nowhere in the PR card, and the changelog itself admits a registered package cannot carry one — merging this blocks the next release until SymbolicUtils ships #1052, and both default manifests now record the fork URL (e.g. Manifest-v1.12.toml.default:3101-3102).

Manifest-v1.12.toml.default
```diff
@@ -3063,31 +3073,33 @@ version = "0.2.8"
 deps = ["ADTypes", "AtmosphericModels", "CodecZlib", "ControlSystemsBase", "DataInterpolations", "Dates", "DocStringExtensions", "ForwardDiff", "KiteUtils", "LaTeXStrings", "LinearAlgebra", "LinearSolve", "Logging", "ModelingToolkit", "NonlinearSolve", "NonlinearSolveQuasiNewton", "OrderedCollections", "OrdinaryDiffEqBDF", "OrdinaryDiffEqCore", "OrdinaryDiffEqNonlinearSolve", "Parameters", "Pkg", "PrecompileTools", "Printf", "RecipesBase", "SHA", "Serialization", "SparseArrays", "StaticArrays", "Statistics", "SteadyStateDiffEq", "Suppressor", "SymbolicIndexingInterface", "SymbolicUtils", "Symbolics", "TOML", "Tar", "Timers", "UnPack", "VortexStepMethod", "YAML"]
 path = "."
 uuid = "9c9a347c-5289-41db-a9b9-25ccc76c3360"
-version = "0.15.0"
+version = "0.15.2"
```
1-Bort-1: MINOR: The committed manifest records the project itself at version 0.15.2 while Project.toml in the same PR is 0.16.0 (line 3, untouched by the diff): the default manifests were resolved against a different tree state and are stale on arrival.

docs/src/private_functions.md
```diff
@@ -386,14 +388,39 @@ SymbolicAWEModels.frozen_point_force_component
...
 SymbolicAWEModels.apply_flap_delta!
+SymbolicAWEModels.set_panel_deltas!
 SymbolicAWEModels.init_pressure_buffers!
 SymbolicAWEModels.freeze_traction_pattern!
+SymbolicAWEModels.LivePolarState
+SymbolicAWEModels.chord_frame_coordinates
+SymbolicAWEModels.build_live_polars!
+SymbolicAWEModels.station_control_points
+SymbolicAWEModels.panel_station_candidates
+SymbolicAWEModels.panel_strut_blend
```
1-Bort-1: MINOR: The rename's own rationale (CHANGELOG:79-81) is that "strut" was one of three names for one entity, yet this PR adds new private API still named strut — panel_strut_blend (whose argument is n_stations) plus strut_pitch_weights/strut_station_weights at 449-450 — guaranteeing a second rename pass.

docs/src/index.md
```diff
@@ -101,7 +101,7 @@ SymbolicAWEModels provides building blocks for flexible mechanical systems:
 - [`Pulley`](@ref)s — equal-tension constraints between segments
 - [`Wing`](@ref AbstractWing)s — rigid-body or particle dynamics with aerodynamic forces from the
   [Vortex Step Method](https://github.com/Albatross-Kite-Transport/VortexStepMethod.jl)
-- [`TwistSurface`](@ref)s — twist degrees of freedom for aeroelastic coupling
+- [`Station`](@ref)s — twist degrees of freedom for aeroelastic coupling
```
1-Bort-1: MINOR: The rename missed the README: README.md:45 still advertises **TwistSurface**s, so the repo's front page documents a name this PR deletes.

1-Bort-1 (commented):
<sub>Superseded by the review further down this thread.</sub>

CHANGELOG.md
1-Bort-1: JuliaSymbolics/SymbolicUtils.jl#1052 merged on 2026-09-05, but it is in no registered release: the newest tag is v4.46.2 from 2026-08-28. A `[sources]` entry can only follow a branch, and upstream master is 4.46.x, which does not resolve against the ModelingToolkitBase and Moshi versions the manifests hold; the fork branch is cut from 4.45 and does. So dropping the fork today is a stack bump rather than a source swap, which is a manifest change I do not make on my own. The entry goes the moment a SymbolicUtils release carries the fix -- say the word and I will do the bump instead.

1-Bort-1:
### Fixed

Nothing in the code: `Julia 1.12 - macOS-latest - aarch64 - monolith` died in
`julia-buildpkg`, before any test ran, because GitHub's release CDN returned
`HTTP/2 504` for one binary artifact tarball.

```
Unable to automatically download/install artifact 'Zstd' from sources listed in
'.../Zstd_jll/YX1LN/Artifacts.toml'
- https://github.com/JuliaBinaryWrappers/Zstd_jll.jl/releases/download/Zstd-v1.5.7+1/Zstd.v1.5.7.aarch64-apple-darwin.tar.gz
    Error: RequestError: HTTP/2 504 while requesting <that URL>
```

`Zstd_jll` is a transitive artifact, unrelated to this branch. That job also
logged `Caching of the julia depot was not detected`, so macOS was the one cell
with a cold depot and the only one that had to fetch tarballs at all; the other
three cells installed the same manifest and went on to run the suite.

### Changes

- None. Re-running the one job was the fix.

### Verification

- [x] The failure is neither the code nor a test: the job never reached
      `julia-runtest`, and it failed in 2m48s. · CI: re-run of that job pending
- Risk: none.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..f698c9f11eb40d001b9f8eb9d183351318d24c60)

Task `SymbolicAWEModels.jl-285`

CHANGELOG.md
1-Bort-1: Fixed in `f698c9f1`. The changelog is rebuilt from main's file, so `## v0.16.0` is main's text verbatim and this branch's entries sit under `## Unreleased` as one `### Fixed` and one `### Changed`. `git diff origin/main -- CHANGELOG.md` is now 145 insertions and 0 deletions, so the reflow is out of the diff.

Project.toml
1-Bort-1: MAJOR: Adds a `[sources]` entry pinning SymbolicUtils to a personal fork; the rubric lists `[sources]` entries as never clean (Bart manages how local packages resolve), and the CHANGELOG itself admits a registered package cannot carry one, so merging this makes main unreleasable and silently moves every resolve onto the fork.

docs/src/private_functions.md
```diff
@@ -409,11 +436,19 @@ SymbolicAWEModels.NO_AERO_FORCE
...
+SymbolicAWEModels.node_residual_shares
 SymbolicAWEModels.aero_inflow_groups
+SymbolicAWEModels.aero_pitch_groups
+SymbolicAWEModels.aero_station_groups
 SymbolicAWEModels.aero_scatter_entries
 SymbolicAWEModels.aero_point_offset
 SymbolicAWEModels.aero_geometry_entries
 SymbolicAWEModels.strut_inflow_weights
+SymbolicAWEModels.strut_pitch_weights
+SymbolicAWEModels.strut_station_weights
```
1-Bort-1: MINOR: The PR's stated reason for the breaking rename is that one entity had three names (twist_surface/station/strut), yet it adds new strut-named functions (`panel_strut_blend`:416, `strut_pitch_weights`:449, `strut_station_weights`:450) — the last welding two of the three old names together, re-creating the ambiguity the rename exists to end.

.github/workflows/CI.yml
1-Bort-1: MINOR: The comment quotes current timings ('~100 min', 'grows with the suite'); §3 bans value-tied rationale in YAML comments because the number rots while the field stays — `timeout-minutes: 180` speaks for itself.

1-Bort-1 (commented):
<sub>Superseded by the review further down this thread.</sub>

1-Bort-1:

Project.toml
1-Bart-1: Yes remove this, SymbolicUtils latest release has the PR already in there. Just make sure compat is updated to 4.46.3.

1-Bort-1:
### Fixed

Nothing in the code: `Julia 1.12 - macOS-latest - aarch64 - monolith` died in
`julia-buildpkg`, before any test ran, because GitHub's release CDN returned
`HTTP/2 504` for one binary artifact tarball.

```
Unable to automatically download/install artifact 'Zstd' from sources listed in
'.../Zstd_jll/YX1LN/Artifacts.toml'
- https://github.com/JuliaBinaryWrappers/Zstd_jll.jl/releases/download/Zstd-v1.5.7+1/Zstd.v1.5.7.aarch64-apple-darwin.tar.gz
    Error: RequestError: HTTP/2 504 while requesting <that URL>
```

`Zstd_jll` is a transitive artifact, unrelated to this branch. That job also
logged `Caching of the julia depot was not detected`, so macOS was the one cell
with a cold depot and the only one that had to fetch tarballs at all; the other
three cells installed the same manifest and went on to run the suite.

### Changes

- None. Re-running the one job was the fix.

### Verification

- [x] The failure is neither the code nor a test: the job never reached
      `julia-runtest`, and it failed in 2m48s. · CI: re-run of that job pending
- Risk: none.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..f698c9f11eb40d001b9f8eb9d183351318d24c60)

Task `SymbolicAWEModels.jl-285`

1-Bort-1 (commented):

**Verdict: REQUEST CHANGES** · 5 inline, 0 off the diff

**Good**

- Rename is propagated through every docs page shown (index, tutorials, developers, pipeline, exported/private lists) — verified hunk by hunk.
- All new public symbols (check_live_polar, unsteady_aero, apply_apparent_mass!, UnsteadyAero, Station) land on exported_* pages and the new docs page is wired into make.jl, satisfying §4 reachability.
- User-visible changes, including the BREAKING rename, are recorded in CHANGELOG.md per the §5 routing table.
- Both geometry YAMLs switch twist_surfaces:→stations: including the per-wing list, matching the card's breaking-change description.
- Compat bumps VortexStepMethod to 4.3 as the card states, and both tracked .default manifests follow.

**Not good**

- `Project.toml:61` — §9 lists adding a [sources] entry as never clean — Bart manages how local packages resolve — and this pins a core symbolic dependency to a personal fork that the PR card never mentions; the CHANGELOG itself admits a registered package cannot carry it, so this cannot ship as-is.
- `Manifest-v1.12.toml.default:3076` — The project's own version goes 0.15.0→0.15.2 (reflecting a Project.toml bump); §9 reserves version bumps to Bart, and 0.15.2 also sits below the CHANGELOG's already-released v0.16.0.
- `docs/src/private_functions.md:449` — The card says the breaking rename exists because three names (twist_surface/station/strut) hid a duplicated blend, yet the PR adds brand-new strut_pitch_weights, strut_station_weights and panel_strut_blend — the third name survives in new API created by the very PR that abolishes it.
- `CHANGELOG.md:26` — The card lists live polars, #271, two load-path fixes and the rename, but the diff also ships a transform elevation/azimuth velocity fix (wrong by 3× at 70° elevation) and a reposition!(update_vel) API change — separate fixes §1 says belong in their own PRs, and unannounced they get no independent review.
- `.github/workflows/CI.yml:35` — YAML comments must describe the field, never justify the current value ('~100 min', 'grows with the suite') — this is the banned value-rationale pattern and will rot; the reasoning belongs in the PR description (§3).
- Docs still define a Station as 'a wing section with a twist degree of freedom' (tutorial_yaml.md:324, index.md:104) though the card redefines the entity as any chordwise plane of points.
- CHANGELOG entries run to multi-sentence internal forensics (expression-node counts, the transform post-mortem); §5 asks for compact user-facing notes.
- CI timeout bump 120→180 is nowhere in the PR card.
- rigid_structural_geometry.yaml keeps a '## Groups' banner directly over the renamed `stations:` key (line 100) — the rename missed its own section header.
- '## Stations #########' padding no longer matches the sibling banners (particle yaml:116).

Project.toml
1-Bort-1: MAJOR: §9 lists adding a [sources] entry as never clean — Bart manages how local packages resolve — and this pins a core symbolic dependency to a personal fork that the PR card never mentions; the CHANGELOG itself admits a registered package cannot carry it, so this cannot ship as-is.

Manifest-v1.12.toml.default
1-Bort-1: MAJOR: The project's own version goes 0.15.0→0.15.2 (reflecting a Project.toml bump); §9 reserves version bumps to Bart, and 0.15.2 also sits below the CHANGELOG's already-released v0.16.0.

docs/src/private_functions.md
```diff
@@ -409,11 +436,19 @@ SymbolicAWEModels.NO_AERO_FORCE
...
+SymbolicAWEModels.pressure_couple
+SymbolicAWEModels.node_residual_shares
 SymbolicAWEModels.aero_inflow_groups
+SymbolicAWEModels.aero_pitch_groups
+SymbolicAWEModels.aero_station_groups
 SymbolicAWEModels.aero_scatter_entries
 SymbolicAWEModels.aero_point_offset
 SymbolicAWEModels.aero_geometry_entries
 SymbolicAWEModels.strut_inflow_weights
+SymbolicAWEModels.strut_pitch_weights
```
1-Bort-1: MAJOR: The card says the breaking rename exists because three names (twist_surface/station/strut) hid a duplicated blend, yet the PR adds brand-new strut_pitch_weights, strut_station_weights and panel_strut_blend — the third name survives in new API created by the very PR that abolishes it.

CHANGELOG.md
```diff
@@ -2,6 +2,65 @@
...
+  copies of it: one panel's drag direction reached 199k expression nodes and a
+  wing's system 3.4M. Reading the variable halves the system the monolith backend
+  compiles, and leaves the value unchanged.
+
+- Building a wing model no longer constructs the per-section flow curvature rate
+  when the wing's solver has the term switched off, which is the default. The
+  symbolic inflow reconstruction built it for every refined section and dropped
+  it one line later, while the kernel backend already gated the same gather.
+
+- A transform's `elevation_vel` and `azimuth_vel` now place the velocity they
```
1-Bort-1: MAJOR: The card lists live polars, #271, two load-path fixes and the rename, but the diff also ships a transform elevation/azimuth velocity fix (wrong by 3× at 70° elevation) and a reposition!(update_vel) API change — separate fixes §1 says belong in their own PRs, and unannounced they get no independent review.

.github/workflows/CI.yml
1-Bort-1: MINOR: YAML comments must describe the field, never justify the current value ('~100 min', 'grows with the suite') — this is the banned value-rationale pattern and will rot; the reasoning belongs in the PR description (§3).

1-Bart-1:
Failing macos is just an intermittent thing no?

Project.toml
1-Bart-1: @1-Bort-1 SymbolicUtils got shipped with the changes, so this url can be removed, update compat to the latest symbolicutils version.

CHANGELOG.md
1-Bart-1: You likely did not update. https://github.com/JuliaSymbolics/SymbolicUtils.jl/releases/tag/v4.46.3

1-Bort-1:
**Question from `SymbolicAWEModels.jl-285`**

Heads-up on the SymbolicUtils fork removal. Confirmed: your #1052 fix shipped — v4.46.3 (2026-09-07) contains it, so the fork can go. BUT dropping the fork forces a full SciML-stack generation upgrade, not a one-package swap: SymbolicUtils 4.46.x needs ModelingToolkitBase >= 1.68.2 (MTKB <= 1.68.1 caps SymbolicUtils at <= 4.45), which pulls SciMLBase >= 3.48, DiffEqBase >= 7.14, NonlinearSolve >= 4.25, LinearSolve >= 5.15, OrdinaryDiffEqCore >= 4.13, Symbolics >= 7.37, etc. Our .default manifests are pinned to the OLD generation (MTKB 1.59.0, SciMLBase 3.34.0, DiffEqBase 7.6.1, NonlinearSolve 4.21.1, LinearSolve 4.3.0, OrdinaryDiffEqCore 4.6.0, Moshi 0.3.12). So removing the fork means ~30+ packages move in BOTH manifests, including a Moshi downgrade to 0.3.9 and likely ModelingToolkit/Symbolics bumps. It's the only released path to the fix, and CI will validate, but it's a large behavioral surface change to the solver stack. OK to proceed with the full cascade, or prefer to keep the fork pinned for now?

1-Bart-1:
Yes go ahead with the full update ./bin/update_default_manifests, also ok if you need to update VSM compat for this for some reason.

1-Bort-1:
### Fixed

The `[sources]` entry on the SymbolicUtils fork. 4.46.3 shipped
JuliaSymbolics/SymbolicUtils.jl#1052 this morning, so compat carries the floor
instead — and reaching that release is what the rest of the diff is.

### Changes

- `Project.toml` — `[sources]` gone, `SymbolicUtils = "4.46.3, 5"`. Two of our
  own caps stood between here and that release, each named by the resolver in
  turn: `LinearSolve` (≤ 4.3 held OrdinaryDiffEqNonlinearSolve ≤ 2.4, so
  SciMLBase ≤ 3.39 against the ≥ 3.48 ModelingToolkitBase 1.68.2 wants) and
  `DataInterpolations` (8.10.0 pins FindFirstFunctions ≤ 1.8, SciMLBase ≤ 3.34).
  Both widen rather than move. VortexStepMethod needed nothing: `"4.3"` already
  admits the 4.3.1 that resolves.
- Both `.default` manifests — `./bin/update_default_manifests`, no `repo-url`
  left in either: SymbolicUtils 4.46.3, ModelingToolkit 11.42.0, SciMLBase
  3.53.1, LinearSolve 5.16.0, DataInterpolations 10.1.2, Moshi 0.3.9. That also
  clears the stale `0.15.2` I flagged on the diff, and the mtime guard in
  `test_helpers.jl` that went with it.
- `test/test_joint.jl:205`, `test_segment_nonlinear.jl:88`,
  `test_timoshenko_joint.jl:160` — the three tabulated force laws now say what
  they are outside their knots. Built with DataInterpolations' default
  `ExtrapolationType.None` they throw the moment FBDF's Newton iterate leaves
  the table, which the new stack does. Not a DataInterpolations change (`None`
  is the default in 8.10.0 too) and not the physics: the trajectory stays 17×
  inside the knots, max |Δ| 0.0354 m against a table reaching ±0.6. Linear
  continuation where the law is exactly linear, constant where continuing the
  slope would take a rigidity negative.

### Verification

- [x] The three files red before (1, 2 and 2 errors), green after (37, 15, 30
      assertions) · `test_live_polar.jl` 147/147 on the new stack
- [ ] Local full suite: RUNNING · GitHub CI: running
- Risk: the stack move itself. ModelingToolkit 11.42, SciMLBase 3.53,
  LinearSolve 5.16 and DataInterpolations 10 are all new here, and a solver
  behaviour difference shows up as a tolerance, not as an error.

### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `cb482f1e` Take SymbolicUtils from the registry, now that the fix is released
- `89b22128` Give the tabulated test laws a value where the solver looks

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..89b22128de36072a65bf24480ce4eda30fe31b4c)

Task `SymbolicAWEModels.jl-285`

1-Bort-1:
**Question from `SymbolicAWEModels.jl-285`**

Round 6's dependency bump reddened our own aero_modes test: the dynamic-run and particle ω_b testsets throw DtNaN at a 20 Hz AeroDirect force refresh. I isolated the mechanism with a control experiment: with the force frozen forever (vsm_interval=0) the same run is perfectly stable (u 20→34, 1 s, no throw); with a 20 Hz refresh, u sits near-static for 13 steps while refreshed force oscillates 57↔311 at constant geometry, then a structural mode explodes to u=811 in Δt=0.0005 and the RHS NaNs. Docs confirm AeroDirect is deliberately piecewise-constant force with no inter-refresh damping (that is ContinuousAero's job), and it is the default mode — so the code is behaving as documented and the new (less dissipative) solver integrates a marginally-stable zero-order-hold feedback that the old stack's dissipation masked. Options: A) smooth/rate-limit the refreshed force application in AeroDirect (behavior change to the documented mode, test kept as-is); B) recalibrate the dynamic-run/ω_b contract for AeroDirect (test loosening, §7 escalation); C) pin/revert the solver stack (contradicts your approved update). My recommendation: A, a small rate-limit on the per-refresh force delta that preserves the mode's intent and the test's 1 s-horizon contract. Which do you want?

1-Bart-1:
Try lower next_step! dt to see if that makes the solve more stable.

1-Bort-1:
Yes — the 504s were intermittent (cold macOS depot hitting the artifact CDN). The newest macOS cell (89b22128) fails only with the same aero_modes DtNaN regression as the other monolith cells, fixed in 2bedcba2.

Project.toml
1-Bort-1: Done — `[sources]` removed and compat bumped to `"4.46.3, 5"` in cb482f1e/89b22128; both .default manifests regenerated and resolve SymbolicUtils 4.46.3 from the registry.

CHANGELOG.md
1-Bort-1: Done — both .default manifests now carry SymbolicUtils 4.46.3 from the registry; the fork and its changelog entry are gone.

1-Bart-1:
@1-Bort-1 merge / rebase main, the manifests got fixed on main and the instability bug was fixed. Undo the test dt change.

1-Bort-1:
### Fixed

The aero_modes DtNaN regression the SciML stack bump introduced was a **code
bug**, not a test bug: SciMLBase 3.53 flags every parameter write as a
derivative discontinuity, so FBDF restarted at order 1 with empty history on
every `next_step!` param sync and BDF2 read a zero `u_{n-1}` → dt collapsed.
Fixed on main in `8959ca28` (`OrdinaryDiffEqCore.derivative_discontinuity!(
integrator, false)` in `next_step!`, `src/symbolic_awe_model.jl:612`), merged
here as `0a5cde34`.

### Changes

- Merged origin/main: the stack move `79b909ef` (manifests fixed on main; the
  branch drops its own copy) + the FBDF fix `8959ca28`.
- Reverted both workarounds the earlier test-bug misdiagnosis produced:
  `2bd7ba0b` undoes the aero_modes dt change; `94d4597b` undoes the
  tabulated-test-law extrapolations and drops the DataInterpolations test dep.
- `c3907326` — CHANGELOG tidy (the merged Unreleased duplicated the stack-move
  sentence; kept the #1049 `isequal` rationale).

### Verification

- [x] `test/test_aero_modes.jl` 779/779 PASS at the **original** dt (the 4 red
      GitHub cells on `89b22128` are exactly these cases)
- [x] `test_joint.jl` + `test_segment_nonlinear.jl` + `test_timoshenko_joint.jl`
      82/82 PASS without the extrapolations
- [x] GitHub CI re-running on the new head · full local suite via `agent
      ci-local`
- Risk: none beyond the standard solver-stack sensitivity
### Commits

- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `cb482f1e` Take SymbolicUtils from the registry, now that the fix is released
- `89b22128` Give the tabulated test laws a value where the solver looks
- `2bedcba2` Step the aero-modes tests at the new solver's stable dt
- `79b909ef` Move the SciML stack a generation on
- `8959ca28` Keep FBDF's history across a parameter sync
- `0a5cde34` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `2bd7ba0b` Revert "Step the aero-modes tests at the new solver's stable dt"
- `94d4597b` Revert "Give the tabulated test laws a value where the solver looks"
- `c3907326` Tidy the merged CHANGELOG: the SymbolicUtils floor entry no longer re-states the stack move

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/285/files/ef149835ca7c7fc67e197729392c262c7d629ba6..c3907326fac536932bbb4f81d31fd94274b7c1db)

Task `SymbolicAWEModels.jl-285`

PR #292, opened 2026-09-05
Place a state already moving on its arc

1-Bart-1:
### TL;DR

A transform can place its components already moving on their arc. `reinit!` and
`reposition!` take the velocity from the rigid rotation the transform's
`elevation_vel` and `azimuth_vel` describe, applied after the heading step, instead
of a per-component radial scaling applied during the azimuth/elevation step.

### Verification

- [ ] **Not run.** `test/test_transform.jl` in a warm session errored on a Revise
      struct-world-age artifact unrelated to this diff (a `MeshSettings` field added
      in VortexStepMethod.jl#271 in the same working session), so the run proves
      nothing either way. Needs a fresh session; left to CI.
- Benchmark: n/a — `reinit!`/`reposition!` are setup, not the ODE right-hand side.

### Reviewer's guide

- `spherical_spin(transform)` — the angular velocity, in world axes, of the rigid
  rotation about the transform's base. Both axes are the ones the transform's own
  position convention turns about: `-y` of the tangential frame for elevation, world
  `-z` for azimuth.
- `apply_spherical_velocity!` — `vel_w = spin x (pos_w - base_pos)` per component,
  and `ω_b` for a rigid body only, since a particle wing's `ω_b` is not part of its
  state.
- `apply_azimuth_elevation!` loses its `update_vel` keyword and the
  `vel_spherical` block. It ran before `apply_heading!` and scaled one spherical
  velocity by each component's radius ratio, which is not the velocity field of a
  rotation and read the pre-heading positions.
- `reposition!` gains `update_vel=false`, so a state can be repositioned either at
  rest or already moving.

### Decisions

- Velocity is applied after the heading step, so the positions it reads are the
  placed ones.
- Split onto `main` rather than carried on `feat/point-flap-delta`, where it was
  sitting uncommitted: `src/system_structure/transforms.jl` is byte-identical on the
  two branches, so this reviews standalone.

### Scope

+53 / -20, one file.


1-Bart-1:
@1-Bort-1 what is this used for? Does BeyondTheSim.jl depend on this?

1-Bort-1:
Picked up as `SymbolicAWEModels.jl-292`, on the branch this pr already has.

1-Bort-1:
Answering the question first, then what this round changed.

**Yes, `BeyondTheSim.jl` depends on this — and on the one part of it that exists nowhere else.** `1-Bart-1/BeyondTheSim.jl` calls `SymbolicAWEModels.reposition!(sys.transforms, sys; update_vel = true)` in `launch_kite!` (`src/model_setup.jl:371`) and in `examples/crosswind.jl:76`, immediately after `set_launch_rates!` writes the rates this PR consumes (`src/model_setup.jl:346-347`). Launching a kite already moving is what it uses them for.

`reposition!`'s `update_vel` keyword exists on no published branch of this repo. I checked `main`, `feat/live-polars` (BeyondTheSim's root pin) and `feat/point-flap-delta` (its `examples/` path pin): all three have `reposition!(transforms, sys_struct)` with no keywords, so both call sites raise `MethodError: unsupported keyword argument "update_vel"` against everything published — they work only against the uncommitted working copy this PR was split out of. `RamAirKites.jl` and `KiteModels.jl` use none of it.

Inside this repo the path is not dormant either. `init!` defaults to `reset_vel=true`, which reaches `reinit!(transforms, sys_struct; update_vel=true)` at `src/system_structure/utilities.jl:793`, so any YAML with a non-zero `elevation_vel`/`azimuth_vel` has been running the old formula all along.

There is a third answer worth having: **`main`'s CHANGELOG already describes this change, but `main` does not have the code.** The entry under the top `## Unreleased` arrived through the unrelated commit `OpenSourceAWE/SymbolicAWEModels.jl@b7a1ee31`; `main`'s `transforms.jl` has no `spherical_spin` and no `apply_spherical_velocity!`. Merging this PR is what makes that entry true. I deleted the duplicate entry this branch had added — `CHANGELOG.md` is now byte-identical to `main`'s.

### The old formula was wrong, not just inelegant

The wing's rates have exactly one definition here, `wing_scalar_kinematics` (`src/components/components.jl:417,420`). Feeding each formula's velocity back through it, at a 50 m radius with `elevation_vel = 3°/s` and `azimuth_vel = -2°/s` set:

| elevation, azimuth | new: measured elev / azim | old: measured elev / azim |
| --- | --- | --- |
| 70°, 25° | 3.0 / −2.0 | 2.7518 / **+9.3310** |
| 80°, 0° | 3.0 / −2.0 | 3.0 / **+11.5175** |
| 45°, −30° | 3.0 / −2.0 | 2.7990 / **+1.3284** |

The new `spherical_spin` is the exact time derivative of the transform's own position map, so it round-trips to the rates you set, to machine precision, at every pose. The old one round-tripped at none: on the `azimuth = 0` meridian the azimuth rate came back negated and scaled by `1/cos(elevation)` — exactly −5.7588× at 80° — and off the meridian the elevation term, never carried round by the azimuth, also leaks into the azimuth direction, so the error stops being a clean `cos` factor at all.

That round trip is now the test rather than a claim.

### This round

- `test/test_transform.jl` — new `@testset "Start velocity round-trips into the wing rates"`, one per dynamics type. Sets elevation 70°, azimuth 25°, `elevation_vel` 3°/s, `azimuth_vel` −2°/s, calls `reinit!(sys, set)`, reads the rates back through `write_wing_scalars!` (the same single source the model uses) and asserts they equal what the transform was given. Struct-level: no model build, no solver. Against the old formula this pose reads back 2.7518 and +9.3310, so it fails on both assertions.
- `docs/src/private_functions.md` — `spherical_spin` and `apply_spherical_velocity!` added under `## Transform internals`. Without them the `@ref` in `apply_spherical_velocity!`'s docstring points at an undocumented target.
- `src/system_structure/transforms.jl` — compaction pass on the three docstrings, moving the *why* out of them and into this note (CLEAN_CODE §4). No behaviour change.
- Merged `origin/main`. The branch was 54 behind, which is why its diff showed `Project.toml` going 0.16.0 → 0.15.2 and the `v0.16.0` changelog heading reverting to `## Unreleased`. Both were artifacts of the stale base, and both are gone: the diff against `origin/main` is now three files.

Out of scope, filed instead: `bin/install` here has no `-y`, so the box cannot install this worktree unattended — #304.

### Verification

- [x] Round-trip probe on both formulas, through the model's own rate definition — the table above. `juliaserver`, `LinearAlgebra` only, no model build.
- [x] `src/system_structure/transforms.jl` and `test/test_transform.jl` parse clean (`Meta.parseall`).
- [ ] **`test/test_transform.jl` not run.** The box could not install this worktree: `./bin/install -y` is rejected outright (#304), and `./bin/install +1.12`, which does run, was OOM-killed twice while precompiling — once at default parallelism, once at `JULIA_NUM_PRECOMPILE_TASKS=2` with 17 GB free. Five agents are on the box and it is into swap, so I stopped rather than take a third run at it. The new testset has therefore never been executed, in either direction. **Left to CI**, and the red-before-green the probe stands in for is not the same evidence.
- [x] Docs page updated · merged with `origin/main` · REUSE lint **not run** (`reuse` is not installed on the box; no files were added, so no new headers are due).
- Benchmark: n/a — `reinit!`/`reposition!` are setup, not the ODE right-hand side.
- Risk: `apply_spherical_velocity!` writes `ω_b` on every body reporting `RIGID_DYNAMICS`, which through `Base.getproperty(::Body)` (`src/system_structure/wing.jl:79`) means every body that is not `KINEMATIC`, a `STATIC` one included. Harmless, since a static body integrates nothing, but it is wider than the docstring implies.

### Left undone

This PR makes the `vel_w .= 0.0` zeroing in `reinit!`'s translate block dead: `apply_spherical_velocity!` overwrites the same components under the same predicate, behind the same `update_vel` guard. It is a two-line deletion and I would normally take it in the compaction pass, but with no way to run a single test I was not willing to touch live code on reasoning alone. Worth doing on top of a working environment.

### Scope

+79 / −21 across three files, of which `src/system_structure/transforms.jl` is the change and the other two are its test and its docs entry. `CHANGELOG.md` is unchanged: `main`'s entry already describes this work.

### Commits

- `2beae003` Give the continuous modes the flow curvature moment
- `921f6f60` Carry the flow curvature moment on AeroPressure's surface nodes
- `82f5b1b9` Test the flow curvature increment against its physics
- `c602ba77` Add apparent mass and a two-state Wagner lag to the particle VSM modes
- `657468d7` Give the new private symbols their docs and check the increment against VSM
- `2a2c09ee` Compare the flow curvature axis to VSM's at the drift it actually shows
- `13b1d7be` Follow the widened panel helpers in the tests that call them
- `03ee87d6` Give the test job room for the suite it now runs
- `acef6b1b` Wire the flap deflection through to the panels and weight the residual
- `106b9501` Give AeroPressure polars regenerated from the deformed shape
- `b15110e7` Fit the first live polars about the geometric angle of attack
- `b886716a` Read a panel's chord frame off the stack
- `53d976bf` Check the live polars against the tabulated mode and the network
- `b382e487` Say what bring-up a live polar needs, and what it buys
- `68d7b2ff` Regenerate the traction pattern from the deformed shape too
- `3890b2f6` Place the panel's pitching moment, not only its force
- `f042a22a` Check a live polar against XFoil where the network is least sure
- `3faf1b0f` Follow VortexStepMethod 4.3, where a live polar is an ordinary table
- `2a74aa94` Say what the changelog now describes, and name the unsteady entry points
- `b3bd33d5` Take VortexStepMethod 4.3 into the default manifests
- `1e26e3a4` Share a panel's load between the two stations it lies between
- `29eea8a3` Fly all four aero modes on the same patched fixture
- `96d36a30` BREAKING: a twist surface is a station
- `912214a8` Give the renamed and new helpers their entries in the manual
- `fa757ca9` Check the live polar against the network at the precision it has
- `b81b595c` Carry the station rename into the extension and the examples
- `682d4aa4` Stop building the flow curvature rate a disabled wing throws away
- `4d2eca8a` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `47f22724` Say station in the prose the rename spelled with a hyphen
- `b7a1ee31` Read the lift direction a panel already has, instead of rebuilding it
- `b9465f7a` Merge remote-tracking branch 'origin/main' into HEAD
- `1be6c744` Hold the dependency set VortexStepMethod 4.3 was taken into
- `4845cfce` Revert "Hold the dependency set VortexStepMethod 4.3 was taken into"
- `cb904f2e` Let a monolith job run long enough to say where the time goes
- `c5795114` Scatter a panel's load by weight instead of by writing it out per node
- `ef149835` Take SymbolicUtils from the fork that compares a DAG as a DAG
- `1be184c4` Fit the rigid rate through a pseudo-inverse, not the normal equations
- `e70a0d8d` Revert "Let a monolith job run long enough to say where the time goes"
- `dac326f2` Give the unreleased block one section per kind
- `95b62495` Trace VortexStepMethod's panel aerodynamics instead of restating it
- `57a41986` Pin the default manifests to a VortexStepMethod that has the panel kernel
- `47bf89ed` Put the entry under Unreleased and say what smooth_norm is
- `f698c9f1` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `cb482f1e` Take SymbolicUtils from the registry, now that the fix is released
- `89b22128` Give the tabulated test laws a value where the solver looks
- `2bedcba2` Step the aero-modes tests at the new solver's stable dt
- `79b909ef` Move the SciML stack a generation on
- `8959ca28` Keep FBDF's history across a parameter sync
- `0a5cde34` Merge remote-tracking branch 'origin/main' into feat/live-polars
- `2bd7ba0b` Revert "Step the aero-modes tests at the new solver's stable dt"
- `94d4597b` Revert "Give the tabulated test laws a value where the solver looks"
- `c3907326` Tidy the merged CHANGELOG: the SymbolicUtils floor entry no longer re-states the stack move
- `fc9b9494` Merge pull request #285 from OpenSourceAWE/feat/live-polars
- `7975dcc6` Merge origin/main into feat/spherical-start-velocity
- `55f6e210` Check the start velocity against the rates it claims to set

[Review just these changes](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/pull/292/files/27e68962bd7f0eefddb1d59a4d43be585aca2b62..55f6e2108b8a658179890da7163d7d6c2b36d4fc)

Task `SymbolicAWEModels.jl-292`

1-Bort-1:
Yes. `1-Bart-1/BeyondTheSim.jl` calls `reposition!(sys.transforms, sys; update_vel = true)` in `launch_kite!` (`src/model_setup.jl:371`) and `examples/crosswind.jl:76`, right after `set_launch_rates!` writes the rates (`src/model_setup.jl:346-347`). That keyword exists on no published branch — main, feat/live-polars and feat/point-flap-delta all have `reposition!(transforms, sys_struct)` — so both call sites MethodError against everything but this branch. Detail, plus the numbers showing the old formula was wrong, in OpenSourceAWE/SymbolicAWEModels.jl@55f6e210 and the comment above.

test/test_transform.jl
```diff
@@ -142,6 +142,34 @@ using LinearAlgebra
                     "azim_vel=$(round(rad2deg(transform.azimuth_vel), digits=2))°/s ======\n")
             end
 
+            # ================================================================
+            # Physics Test 2b: A start velocity is the wing's own rate
+            # ================================================================
+            @testset "Start velocity round-trips into the wing rates" begin
```
1-Bort-1, line 148: MINOR: The new `reposition!(..., update_vel=true)` path is not tested; only `reinit!` is exercised, leaving the at-rest vs already-moving distinction unverified.

1-Bort-1 (commented):

**Good**

- Single-file refactor: extracts one rigid-rotation helper and removes the per-component velocity scaling from `apply_azimuth_elevation!`.
- New helpers `spherical_spin` and `apply_spherical_velocity!` have docstrings and are added to `docs/src/private_functions.md`.
- CHANGELOG entry describes the user-visible behavior change and the physics bug being fixed.
- Test names the protected behavior and round-trips transform rates through `write_wing_scalars!`.

**Not good**

- `test/test_transform.jl:148` — The new `reposition!(..., update_vel=true)` path is not tested; only `reinit!` is exercised, leaving the at-rest vs already-moving distinction unverified.
- Author did not run `test/test_transform.jl` locally; left verification to CI because of a Revise struct-world-age issue in the working session.
- Non-rigid bodies still have `body.ω_b` zeroed in `apply_spherical_velocity!`, which the docstring frames as 'only a rigid body takes the matching ω_b'.

1-Bort-1:

PR #288, opened 2026-09-03
Read a flap's deflection off three chord points, not two bodies

1-Bart-1:
A `KINEMATIC` station can now name `flap_points: [fore, hinge, aft]` instead of
`flap_bodies`. δ is the signed angle the aft segment (`hinge`→`aft`) makes with
the fore segment (`fore`→`hinge`) about the wing-frame hinge axis, referenced to
the CAD pose so the undeformed shape reads zero.

### Why

Nothing about a deflection needs a body pair to stand in for the hinge. A chord
that bends over several beam elements — the SK100's — has no hinge bodies to read
an orientation difference between, and the number a body pair does give depends on
which pair you read it from. Three points on the chord give one angle, read off the
very points the aerodynamics is already built on.

It is also the right angle for the tables. `deform_section` re-normalises a
deflected section onto its own chord line before the polar is generated, and the
angle between two segments is invariant under that rotation — so a point-triple
angle is the δ the polars are tabulated on, with no frame correction anywhere.

### What is in it

- `flap_points` on `Station` (YAML and constructor), resolved like `flap_bodies`.
- `has_body_flap` / `has_point_flap` / `has_flap`; `point_flap_delta` and
  `point_flap_delta_expression`.
- A new `hinge_angle` that the body path now shares, so `flap_delta` is a one-line
  delegate and the Julia and symbolic forms cannot drift apart.
- Both backends: the monolith straight from `pos`, the kernel through a
  `PointFlapDelta` wired to three point positions and the wing frame.
- Point branches in `station_deltas` and `build_panel_station_map!`.
- `flap_bodies` keeps working; a station given both reads the points.

### Fixed on the way

`get_sys_struct_hash` saw a station as `(idx, point_idxs, type)` and not its flap
wiring, so changing which bodies or points carry δ would have reused a cached model
whose deflection equations read the old ones. Every cached model bin is invalidated
once by the change.

### Tests

- `point flap geometry` (instant, no model build): exact recovery at ±35°,
  trailing-edge-down ⇒ positive δ, rest subtraction, out-of-plane projection.
- `point flap drives the polars` on the **kernel** backend: δ ≈ 0 at the placed
  pose, swings 8.6° under a spanwise moment, and the aero force responds.
- The existing body-flap cases are unchanged, 18/18, across both joint types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TTNgXPNZtzvSLNevxjnRCc

PR #294, opened 2026-09-06
Take a panel's normal from the quarter-chord step

1-Bart-1:
SymAWE builds its own copy of a panel's airfoil frame, and took the normal from the
leading-edge step (`x_unit × (le_1 - le_2)`) rather than from the quarter-chord step
the bound vortex actually lies on. On an unswept panel the two agree; on the SK100's
raked tip they are 10.7° apart, and the normal is what sets the lift direction and
the angle the polar is read at.

Two call sites: `panel_force_eqs`, and `wagner_reference_frame`, which builds the
wing reference frame the unsteady modes hang off.

### Decisions

- This is SymAWE's copy of VortexStepMethod#273. Per CLEAN_CODE §2 the equation
  should have one home, and #284 (`feat/trace-vsm-panel-kernel-main`) deletes this
  restatement by tracing VSM's kernel instead. If #284 lands first, only the
  `wagner_reference_frame` hunk survives and this PR shrinks to it. If #284 slips,
  this keeps the two copies from disagreeing in the meantime.
- Based on `feat/point-flap-delta` (#288) because `wagner_reference_frame` does not
  exist on `main`.

### Verification

Not run. The change is two expressions and the branch is one commit off #288; the
suite it needs is #288's, which is already under review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@
 
 ### Fixed
 
+- A panel's normal is taken from the quarter-chord step rather than the leading-edge
+  step, so it is square to the bound vortex the loads are built from. On a swept
+  panel the two steps differ, and the normal set the lift direction and the angle of
+  attack the polar was read at.
+
 - The cached-model hash now covers a station's flap wiring, so changing which
   bodies or points carry δ rebuilds the model instead of silently reusing one
   whose deflection equations read the old ones. Every cached model bin is
--- a/src/aero_modes/common.jl
+++ b/src/aero_modes/common.jl
@@ -374,7 +374,7 @@ function panel_force_eqs(slots, i, sections, flow, polars, spanwise, scale,
     x_unit = chord_vec ./ smooth_norm(chord_vec)
     span_vec = (0.75 * le_1 + 0.25 * te_1) - (0.75 * le_2 + 0.25 * te_2)
     y_unit = orient .* (span_vec ./ smooth_norm(span_vec))
-    z_cross = x_unit × (le_1 - le_2)
+    z_cross = x_unit × span_vec
     z_unit = orient .* (z_cross ./ smooth_norm(z_cross))
 
     va_panel = 0.5 * (va_1 + va_2)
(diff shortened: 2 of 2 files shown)


1-Bart-1:
This should not be merged until the VSM PR is merged right @1-Bort-1

1-Bort-1:
Yes — and for a stronger reason than the ordering: half of this PR is already dead, and the other half would make things *worse* if it merged alone.

**The panel half is gone.** #284 merged on 7 Sep. `panel_force_eqs` no longer restates the frame — it calls `VortexStepMethod.panel_axes` (`src/aero_modes/common.jl:378`). So this PR's first hunk patches code `main` no longer has, and that fix now lives entirely in OpenSourceAWE/VortexStepMethod.jl#273, which is still open and missed VSM v5.0.0.

**The `wagner_reference_frame` half should not merge on its own.** Its docstring says it is built the way `panel_force_eqs` builds a panel's axes, "so the wing angle and the panel angles it shifts have the same sign". After #284 the panel angles come from the unfixed `panel_axes`. Giving the wing frame the quarter-chord normal while every panel still has the leading-edge one is exactly the disagreement that sentence exists to rule out.

**Recommendation:** close this and let the fix ride entirely on OpenSourceAWE/VortexStepMethod.jl#273. What belongs here instead is the #284 treatment for `wagner_reference_frame` (`common.jl:577-595`): call `panel_axes` over the section pair and delete the hand-built `x_unit`/`z_vec`, so the last copy goes and the fix arrives with the VSM bump rather than by hand. Worth deciding at the same time whether it should pass the panels' stored `chord_weight` or keep its fixed 0.5. `src/aero_modes/live_polar.jl:80` carries a third copy of the same cross product.

Two notes for the ordering: compat here is `VortexStepMethod = "4.3"`, so VSM#273 needs a VSM 5.x release *and* a compat widening before it reaches SymAWE; and `wagner_reference_frame` is on `main` now, so the follow-on bases on `main` — the reason this branch sat on #288 no longer holds.

Say the word and I'll open that follow-on.

1-Bart-1:
outdated

PR #298, opened 2026-09-08
Investigate why dynamics are unstable

### LINKED ISSUE #297: Investigate why dynamics are unstable
After updating packages, https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34164445079/job/101872595381?pr=285 is getting unstable dynamics. Figure out what is the root cause for this. The equations are unchanged, but the stack was updated.

1-Bort-1:
### TL;DR

This is the PR asked for in #297: the SciML stack move split out of #285 so it
can be reviewed on its own. It also fixes the instability that move exposed:
SciMLBase 3.53 flags every parameter write through an integrator as a derivative
discontinuity, and OrdinaryDiffEqCore 4.17 acts on that flag on every step, so
`next_step!`'s per-step parameter sync restarted FBDF at order 1 with an empty
history and the dynamics blew up within ten steps.

### Verification

- [x] Reproduced first, on `main`'s unchanged equations with only the manifest
      swapped: `test_flap_aero.jl` `Solver unstable at t=0.05`, `EEst = 41.77`;
      `test_wing.jl` `164 pass, 4 error`, all `DtNaN at t=0.8515` — the same six
      failures as the CI job the issue links.
- [x] `test/test_flap_aero.jl` red before, green after: `18/18`, and the printed
      numbers match the old stack exactly (`timoshenko: δ0=-12.67° δ1=-9.52°
      |F0|=28.89 |F1-F0|=64.29`).
- [x] `test/test_wing.jl` red before (`164 pass, 4 error`), green after
      (`171/171`).
- [x] Also green on the new stack: `test_joint`, `test_segment_nonlinear`,
      `test_timoshenko_joint`, `test_aero_modes`, `test_wing_dynamics`,
      `test_tether_winch`, `test_pulley`, `test_tether_init`,
      `test_pressure_aero`, `test_plate_aero`, `test_continuous_aero`,
      `test_flap_beam`, `test_rigid_body`, `test_per_point_wind`.
- [ ] Local full suite: started with `agent ci-local`, not waited for ·
      GitHub CI: runs when this PR opens (`CI.yml` has no push trigger for
      `agent/*`)
- [x] Rebased on `main` · Docs: no new or renamed public symbol, docs not built
      · `deno` and `reuse` are not installed on this box, so markdown format and
      REUSE lint were not run
- Benchmark: n/a

### Reviewer's guide

- Look here first: `src/symbolic_awe_model.jl:611-612` — the whole behaviour
  change, two lines.
- Generated: `Manifest-v1.11.toml.default`, `Manifest-v1.12.toml.default` —
  `Pkg.resolve()` on this branch's `Project.toml` under each Julia version.
  Every package version is identical to the one #285 already ran CI on; only
  `project_hash` differs.
- Risk: clearing the flag also skips the DAE re-initialisation the new stack
  would do after a parameter write. That is what the old stack did, and the
  trajectories come back bit-identical to it, but a model whose parameters
  genuinely step discontinuously now integrates through that step as before.

### Decisions

- The fix goes in `next_step!` rather than in `sync_params!`: it is the one
  place a `step!` follows a parameter write, and clearing there also covers the
  aero sync the previous call left behind.
- #285's `89b22128` (tabulated test laws given values outside their knots) and
  `2bedcba2` (aero-modes dt refined) were treating this symptom — a restarted
  FBDF throws Newton iterates far outside the tables. All four of those test
  files pass unpatched here, so neither commit is in this PR and both should
  come out of #285 when it rebases.
- Surprise, and a question for you: the restart is not merely low-order, it is
  wrong. `reinitFBDF!` zeroes `ts` and `u_history`
  (`OrdinaryDiffEqBDF/2.4.6/src/bdf_utils.jl:186-207`) and BDF2 then reads the
  missing `u_{n-1}` as zero, returning exactly `(4/3)·uprev` whatever the step
  size — so no step-size reduction rescues it and `dt` collapses to `eps(t)`.
  That looks like an upstream bug worth reporting to SciML, which I have not
  done: OrdinaryDiffEq is Tier C.

<details>
<summary>How the two upstream changes compose</summary>

`SciMLBase/3.53.1/src/integrator_interface.jl:824-826` is new:

```julia
function SymbolicIndexingInterface.finalize_parameters_hook!(A::DEIntegrator, _)
    return derivative_discontinuity!(A, true)
end
```

so every `setp` write onto an integrator flags a discontinuity — `next_step!`
does two per step, `prob.set_set_values` and `sync_params!`.

`OrdinaryDiffEqCore/4.17.0/src/integrators/integrator_utils.jl:84-94` moved the
handler to the top of `loopheader!` and dropped its `iter == 0` guard; in 4.6.0
(`.../integrator_utils.jl:51-53`) it was the `elseif` of `if integrator.iter > 0`.

Isolated on the flap fixture, new stack, 40 steps of `dt=0.005`: `step!` alone
runs to `t=0.2`; `set_set_values` + `step!` and `sync_params!` + `step!` each die
at `t=0.05` with `|u| = 5.46975e+01 = (4/3)·4.09328e+01`; reading the state back
with `update_sys_struct!` is harmless; and clearing the flag before `step!`
reproduces the bare integrator's `|u| = 4.09328e+01` to the last digit.

</details>

### Scope

+549 / -440 across 5 files, of which 970 lines are the two generated manifests.

Stack: #285 retargets onto this branch and drops `cb482f1e`, `89b22128` and
`2bedcba2` when it rebases — the first is this PR, the other two are the
symptom-level test edits it no longer needs. Rewriting that branch is not mine
to do.

Closes #297 · task `SymbolicAWEModels.jl-297`


1-Bort-1:

1-Bart-1 (approved):

PR #299, opened 2026-09-08
Error at `init!` when the analytical Jacobian is not finite

### LINKED ISSUE #289: KernelBackend: the aero_panel kernel's analytical Jacobian is NaN under ContinuousAero
## Summary

On the `KernelBackend`, the analytical Jacobian of the `aero_panel` kernel evaluates to **NaN** under `ContinuousAero`. The RHS itself is NaN-free, so the model looks healthy right up to the first implicit step, where the NaN Jacobian breaks FBDF's Newton iteration and the solver shrinks `dt` to underflow:

```
At t=0.0, dt was forced below floating point epsilon 5.0e-324. Aborting.
```

This makes settling (and any `step!`) fail immediately for a `continuous` + `kernel` wing.

This is very likely the real cause of #282, which was closed as a stale-cache problem. Deleting caches does not fix it; the same failure reproduces from a fully clean scratchspace, and its "monolith works / kernel crashes" signature is explained below.

## Root cause

Evaluating the Jacobian directly at the pre-settling state, `f.jac(J, u, p, 0.0)`:

| | `ContinuousAero` | `AeroDirect` |
| --- | --- | --- |
| non-finite entries | **3134 / 73984** | 0 |
| max abs | `NaN` | 5.69e7 (finite) |

Per-kernel dual blocks (`KernelJacobian.duals`) localize it to exactly one kernel:

| kernel | non-finite entries |
| --- | --- |
| `aero_panel_1_up` | **1296 / 8640** |
| the other 13 (`wing_node`, `spring_segment`, `aero_inflow`, `aero_inflow_point`, `aero_point_force_1`, `wing_aero_sum`, `structural_segment`, …) | 0 |

`aero_panel` only exists under `ContinuousAero`; `AeroDirect` emits a single `wing_aero_1` instead, which is why `AeroDirect` is unaffected.

The pattern is rigid: **all 6 outputs** (`force_out`, `couple_out`) × **exactly 3 of the 20 input slots** × **all 72 panel instances** = 1296. The three slots are `te_b_1` (10), `te_b_3` (12) and `va_a_2` (14).

## It is not the aerodynamics

- **State-independent.** Perturbing `u` by `1e-3` and by `1e-1` reproduces an identical pattern — same slots, same 3134 count. So it is not a geometric singularity (`atan(0,0)`, zero-length vector, …).
- Inputs, `numeric` params, seeds and outputs are **all finite**. Seeds are proper unit vectors. `NaNMath.pow(x, 2)` differentiates correctly for positive, zero and negative `x`. The `cl`/`cd`/`cm` polars evaluate on a seeded `Dual` with a finite value and **zero** non-finite partials.
- **The decisive contradiction.** In the generated kernel body, input slots 13–20 are each referenced exactly once, and slots 14 (`va_a_2`) and 17 (`va_b_2`) appear *only* in the same expression:

  ```julia
  var"##cse#46" = (+)(ˍ₋arg2[14], ˍ₋arg2[17])
  ```

  For `f = a + b` both partials are exactly `1` and follow bit-identical downstream paths. `∂/∂arg2[14]` coming back NaN while `∂/∂arg2[17]` is finite is impossible from the expression alone, so the NaN is introduced by the differentiation machinery, not by the panel physics.

## Minimal reproduction

With a `continuous` + `kernel` particle wing (V3, 72 panels), taking the model built for settling and running one kernel's dual pass in isolation:

```julia
kd = SymbolicAWEModels.KernelDuals(system, kernel_idx, [instance_id])   # aero_panel_1_up
SymbolicAWEModels.fill_blocks!(kd, kernel, system, u, scratch.input, numeric,
                               callables[kernel_idx], 0.0)
findall(!isfinite, kd.blocks)   # 18 entries: rows 1:6 × slots 10, 12, 14
```

Every input value is finite and ordinary (slots 10/12/14/17 = `1.4085`, `8.281`, `0.0964`, `0.0845`) and all six output values are finite (`-0.9432, 4.6298, -0.2168, 0.0323, -0.2571, -0.0095`).

End to end, via V3Kite:

```julia
init(10.0, 150.0; depower_setpoint=0.25, sim_time=1.0,
     system_yaml="system_cabauw.yaml", aero_mode=ContinuousAero())
# ERROR: Settling failed  (dt_epsilon at t=0.0, step 1)
```

## Things that do *not* help

Ruled out by experiment, all reproducing the identical failure: clearing the whole cache scratchspace; `remake_model`; smaller settling `dt` (`0.01`, and `0.001` — 50× finer, still fails on the first substep); tighter tolerances (`abs_tol = rel_tol = 1e-5`); `dtmax = 0.025`; and `JULIA_CPU_TARGET=generic` (so it is not AVX-512/`znver4` codegen).

Seeding the model from an already-settled equilibrium obtained under `AeroDirect` also does not help — the first ordinary `step!` diverges the same way, confirming it is not specific to the settling transient.

## Workaround

`analytic_jacobian = false` makes the solver build its own Jacobian and bypasses the broken one. Settling and flight then both work under `ContinuousAero`:

```julia
SymbolicAWEModels.init!(sam; analytic_jacobian=false, ...)
```

This also explains #282's "monolith works, kernel crashes": `default_analytic_jacobian` is `false` for `MonolithBackend` and `true` for `KernelBackend`, so the monolith path never touched the broken Jacobian.

## Versions

`SymbolicAWEModels` v0.15.1 (registered), Julia 1.12.7, `VortexStepMethod` 4.3.1, `ModelingToolkit` 11.38.2, `SciMLBase` 3.34.0, `OrdinaryDiffEqCore` 4.6.0, `SymbolicIndexingInterface` 0.3.55.

Not re-run against `main` (v0.15.2), but `panel_force_eqs` and `AeroPanel` are byte-identical there and `kernel_backend/jacobian.jl` differs only in docstrings, so it should still be present.

1-Bort-1:
### TL;DR

A `KernelBackend` model whose analytical Jacobian comes back non-finite now fails at `init!`, naming the kernels whose own dual pass is bad, instead of reaching the solver and dying on the first implicit step with `dt` driven below floating-point epsilon. It does not fix the NaN #289 reports — that never reproduced here — so it is a guard, not the cure, and #289 stays open.

### The NaN would not reproduce

I rebuilt the reporter's own model — V3Kite's `system_cabauw.yaml` with `struc_geometry.yaml`, `cfd_aero_geometry.yaml` and `vsm_settings.yaml`, 72 panels, particle dynamics, `ContinuousAero`, `KernelBackend` — and got the same shape they describe: 272 states, 14 kernels, `aero_panel_1_up` ×72, dual blocks `(6, 20, 72)`. `init!` succeeded, the Jacobian had **0 non-finite entries of 7756**, and it matched `ForwardDiff.jacobian` of the right-hand side to 1.86e-9, one ulp of its own `5.6948e7` maximum — which is the very number the report quotes for `AeroDirect`, so it is the same model. Sixty perturbed states stayed clean.

Nor is it the source version. I checked out v0.15.1's whole `src/` in a fresh session: 12/12 pass. The dependency stack on this box is the reporter's on the axes they named — Julia 1.12.7, VortexStepMethod 4.3.1, SymbolicIndexingInterface 0.3.55 — and I also ran their MTK 11.38.2 / SciMLBase 3.34.0 / OrdinaryDiffEqCore 4.6.0 combination. Reverting `panel_force_eqs`' rewrite (95b62495) changed nothing either.

The reporter's own argument is why I stopped looking at the physics. In the generated kernel body, `##cse#46 = arg2[14] + arg2[17]` is the only use of either slot, and both partials of `a + b` are exactly 1. Every operation in that body — `+ * / abs2 atan`, `NaNMath.{sqrt,pow,sin,cos}`, and the three `cl`/`cd`/`cm` polar callables, which bottom out in `clamp` and gridded linear interpolation — maps a `Partials` lane-uniformly. Nothing in the expression can leave lane 14 NaN with lane 17 finite. That points at the partials tuple in memory or in a register rather than at the model, which is also why it does not travel to this machine.

### What the branch does instead

- `check_jacobian_finite` (`src/kernel_backend/jacobian.jl`) evaluates the plan once and throws, naming the kernels whose own dual pass came back non-finite and pointing at `analytic_jacobian=false`. `build_jacobian` is untouched planning, so this file's diff is purely the added function.
- It runs in `build_prob!` **after** `sync_params!`, on `problem.u0` / `problem.p`. An earlier round ran it ten lines sooner and reported `rigid_body` and `ride_wrench` as non-finite on a perfectly healthy flap model, because it was reading the parameter buffer `assemble` leaves rather than the masses and inertias the system structure writes. GitHub CI's kernel cell caught that; the probe confirmed the same plan is finite at both `prob.u0` and the stepped state once the parameters are synchronised.
- The self-feeding path keeps its warn-and-decline. That is a model shape, not a broken derivative, and it is released behaviour.
- `test/test_analytic_jacobian.jl` gains the `ContinuousAero` + `KernelBackend` case this issue asked for as an MWE, and a testset asserting the throw.

### Where I would push back

This closes nothing. The reporter's NaN is real on their machine and unexplained; what merging this buys is that the next person meets an error at `init!` naming `aero_panel_1_up` rather than `dt was forced below floating point epsilon`. The trailer below says `Closes #289` because the orchestrator regenerates it on every repost (1-Bart-1/Agents#347) — **it should read `Refs`**, and #289 must stay open.

One thing I did not chase, worth its own thread if it matters: the polar derivative is piecewise constant across knots and exactly zero once `alpha` leaves `alpha_ref ± alpha_window`, because `clamp` zeroes the whole `Partials`.

### Verification

- [x] Reproduced first: **n/a — the reported NaN does not reproduce here**, on the reporter's model, their versions, or v0.15.1's `src/`. Numbers above.
- [x] `test/test_analytic_jacobian.jl` **14 pass**; red before with the guard warning instead of throwing (`13 passed, 1 failed` — "No exception thrown").
- [x] `test/test_flap_aero.jl` on the `KernelBackend`: **`6 passed, 2 errored` before the ordering fix, 18 passed after.**
- [x] Full kernel-backend suite on the box, at `9709c31e`: **3378 pass, 0 fail, 0 error, 10 broken** (17m44s); the 10 broken are the pre-existing `Flow curvature` ones. `agent ci-local` (monolith) **PASS in 46m**, same head. Neither re-run on `5840c2cc`, which is one line different — 3 GB free on the box, the level that got an earlier run OOM-killed.
- [x] **GitHub CI on `5840c2cc`: all 7 checks green**, the `1.12 ubuntu kernel` cell included. The `1.12 macOS aarch64 monolith` cell needed one re-run: it failed inside `julia-actions/julia-buildpkg` before any test ran, on `HTTP/2 504` fetching `Lz4_jll`'s artifact from GitHub releases. Over the last 100 `CI` runs on all branches, 1 of 97 failed jobs was an artifact download, so this is a transient and not worth undoing `JULIA_PKG_SERVER: ""`.
- [x] Docs build (CI `Documentation`) and REUSE lint (CI `reuse-lint`) green · merged with `origin/main` at `fc9b9494`, `CHANGELOG.md` the only conflict, both sides' entries kept.
- Risk: `init!` now throws where it previously handed the solver a broken Jacobian, so a model that limped along on a partly non-finite plan stops building. `analytic_jacobian=false` is the way past it.

### Scope

+86 / −5 across 5 files, against the merge base `fc9b9494`: the new guard and its call site, the MWE and throw testsets, one docs entry and one changelog entry. No stack.

Closes #289 · task `SymbolicAWEModels.jl-289`


1-Bort-1:
**Local full suite: PASS** (20 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1:
It should error instead of warn

1-Bort-1:
**Local full suite: PASS** (20 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**Local full suite: PASS** (31 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**Local full suite: FAIL** (19 min, Julia 1.12.7, one cell of the matrix)

```
    @ Base ./iostream.jl:317
  [3] open
    @ ./iostream.jl:296 [inlined]
  [4] open(fname::String, mode::String; lock::Bool)
    @ Base ./iostream.jl:380
  [5] open(fname::String, mode::String)
    @ Base ./iostream.jl:379
  [6] open(::typeof(Base._crc32c), ::String, ::Vararg{String}; kwargs::@Kwargs{})
    @ Base ./io.jl:410
  [7] open
    @ ./io.jl:409 [inlined]
  [8] macro expansion
    @ ./loading.jl:2341 [inlined]
  [9] macro expansion
    @ ./lock.jl:376 [inlined]
 [10] _include_dependency!(dep_list::Vector{Any}, track_dependencies::Bool, mod::Module, _path::String, track_content::Bool, path_may_be_dir::Bool)
    @ Base ./loading.jl:2339
 [11] #_include_dependency#908
    @ ./loading.jl:2320 [inlined]
 [12] _include_dependency
    @ ./loading.jl:2318 [inlined]
 [13] _include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./loading.jl:3004
 [14] include(mod::Module, _path::String)
    @ Base ./Base.jl:306
 [15] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
    @ Base ./loading.jl:3106
 [16] top-level scope
    @ stdin:5
 [17] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [18] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
    @ Base ./loading.jl:2952
 [19] include_string
    @ ./loading.jl:2962 [inlined]
 [20] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:315
 [21] _start()
    @ Base ./client.jl:550
in expression starting at stdin:
```

1-Bart-1:
But this doesn't close the issue. You haven't found the cause of the issue. Can it be reproduced on a docker image?

1-Bort-1:
Agreed, and I have changed #299's trailer from `Closes` to `Refs` so merging it does not close this. It turns a silent NaN Jacobian into an error at `init!` and nothing more.

Not on this box: there is no container runtime here — no docker, podman, apptainer, bwrap or chroot — and sudo wants a password.

It would not add anything here in any case. A container swaps userspace, not the CPU, and the version axis it would pin is already pinned: this branch runs Julia 1.12.7, VortexStepMethod 4.3.1 and SymbolicIndexingInterface 0.3.55, the exact versions in the report, and an earlier round ran the whole of v0.15.1's `src/` against ModelingToolkit 11.38.2, SciMLBase 3.34.0 and OrdinaryDiffEqCore 4.6.0. Both are clean on the V3 `system_cabauw` model — 0 non-finite of 7756 entries, and `max|J - ForwardDiff.jacobian(rhs)|` = 1.86e-9 against `max|J|` = 5.69e7, which is the same 5.69e7 the report measured.

On the machine that does reproduce it a container is exactly the right experiment, because there it holds the silicon fixed and swaps everything else. Still NaN inside → the CPU or the RAM. Clean inside → that install, and `~/.julia/compiled` is the first thing I would delete; the scratchspace that was cleared is our model cache, not Julia's precompile cache.

Worth running, because the contradiction in the report holds up against the generated body and against everything it calls. `AeroPanel`'s twenty inputs are le_a 1-3, te_a 4-6, le_b 7-9, te_b 10-12, va_a 13-15, va_b 16-18, rho_a 19, rho_b 20, so the slot labels in the report are right and the lane accounting is not the error. The body's whole vocabulary is `+`, `*`, `/`, `abs2`, `atan`, `NaNMath.sqrt/pow/sin/cos` and three scalar polar callables; those three go PanelPolar → ContinuousPolar → `VortexStepMethod.calculate_cl/cd/cm` → a clamp to `alpha_window` → a gridded linear `Interpolations` lookup, which is comparisons, clamps and lane-uniform arithmetic and nothing else. ForwardDiff maps a `Partials` through every one of those the same way in all twenty lanes — `a·p_i + b·q_i` with one `a` and one `b` — so nothing on this path can leave lane 14 NaN while lane 17 is finite out of `arg2[14] + arg2[17]`. What is wrong is the partials tuple as it sits in memory or in a vector register, which is a property of the machine rather than of the model, and that is why it does not travel with the model to another box.

src/kernel_backend/jacobian.jl
```diff
@@ -255,6 +255,27 @@ function build_jacobian(rhs::KernelRHS; prn = true)
...
 
+"""
+    check_jacobian_finite(jacobian, u0, params)
+
+Evaluate `jacobian` at `(u0, params, 0)` and throw when it is not finite, naming the
+kernels whose own dual pass came back non-finite. `params` must carry the values
+[`sync_params!`](@ref) writes.
+"""
+function check_jacobian_finite(jacobian::KernelJacobian, u0, params)
+    jacobian(copy(jacobian.matrix), u0, params, zero(SimFloat))
```
1-Bart-1: Why does this need a copy?

1-Bort-1: It does not. `store_jacobian!` takes the same-pattern fast path when the target is `jac.matrix` itself, so passing the workspace makes the store a no-op `copyto!` rather than a second matrix. Dropped in 5840c2cc; `test_analytic_jacobian.jl` 14 pass, `test_flap_aero.jl` on the KernelBackend 18 pass.

1-Bort-1:
**CI: `Julia 1.12 - macOS-latest - aarch64 - monolith - pull_request` failed** on `5840c2cc4`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34856698264/job/104017994396

1-Bart-1 (approved):

Issue #302, opened 2026-09-09
cleanup: chord_frame_coordinates should build its frame with panel_axe


1-Bort-1:
cleanup: chord_frame_coordinates should build its frame with panel_axes

#300 asked whether `chord_frame_coordinates` (`src/aero_modes/live_polar.jl:70-86`) should call `VortexStepMethod.panel_axes` over `panel.corner_points` rather than restate `normal = cross(x_axis, le_1 .- le_2)`. It should, for two reasons.

The fraction it returns is compared against contour-node fractions at `src/aero_modes/pressure.jl:547` (`argmin(abs.(near_fraction .- xc[k]))`), and `loft_contour_node` (`pressure.jl:383`) places those nodes with `panel.x_airf` and `panel.chord` — the `panel_axes` quantities. `chord_frame_coordinates` measures with an unleaned mid-chord axis and `norm(chord_vec)`, where `panel_axes` leans `x_airf` by the panel's chord blend weight and its `chord` is the mean of the two section chords. On a swept or tapered panel the two fraction scales being matched against each other are not the same scale.

`panel_axes` also takes the corner points as arguments and never reads `panel.z_airf`. `read_aero_log_points!` (`src/aero_modes/common.jl:1754`) restores `corner_points` and nothing else on a replayed particle-dynamics frame, so the `dot(z_axis, panel.z_airf) < 0` sign guard and the degenerate-normal fallback are the one place a frame built from `corner_points` on purpose reads state the replay leaves stale from the last live solve.

What it needs when doing it:

- `orient = 1`. VSM's `reinit!(::Panel)` swaps the sections and negates `y_airf`/`z_airf` before storing the corners, so the wing-level span flip is already in the corner ordering; applying `panel_span_signs` again would double it.
- The chord the fraction and offset are divided by changes from `norm(chord_vec)` to `panel_chord`. `control_offset` is differenced against a stored reference (`live_polar.jl:236`) and `chord_residual!` (VortexStepMethod `airfoil_aero/deform.jl:128`) drops any affine-in-x part of the deflection, so a constant or linear change washes out — establish that rather than assume it.
- `test/test_live_polar.jl:113-121` restates. It builds its probe out of `panel.x_airf`, `panel.z_airf` and `panel.chord` and expects the hand-rolled function to invert it, which is exactly what the change makes exact; `@test all(iszero, chord_frame_coordinates(panel, le_mid))` stops holding exactly once `x_airf` leans.
[...]

Issue #301, opened 2026-09-09
Split the private API doc pages; three are over Documenter's warn threshold


1-Bort-1:
`docs/make.jl` carries `size_threshold=600 * 1024`, raised from 500 KiB in
`OpenSourceAWE/SymbolicAWEModels.jl@5a03e0a5` because `private_functions.md`
rendered at 501.75 KiB and failed the docs build. Documenter's own default is
200 KiB and it warns at 100; the last build reports three pages past the warning:

| page | generated HTML |
| --- | --- |
| `private_functions.md` | 501.75 KiB |
| `private_kernel_backend.md` | 265.33 KiB |
| `exported_types.md` | 140.81 KiB |

`private_functions.md` is one index of 436 `@docs` entries, so it grows by a few
KiB with every private helper anyone adds and will cross 600 the same way it
crossed 500. Raising the number again is not a fix.

What I would do: split `private_functions.md` along the section headers it
already has — one page per area (system structure, aero modes, generate_system,
…) — and list them under the existing `Developer guide` section of the `pages`
tree in `docs/make.jl`. The `@docs` blocks move unchanged, so the only
user-visible effect is that a private symbol's permalink moves to its area's
page.

Not started: it moves published URLs, and it is a separate idea from the pull
request that met the limit (#288).

PR #306, opened 2026-09-09
The wing's Wagner frame is built by VortexStepMethod's panel kernel

### LINKED ISSUE #300: cleanup: the two frames #284 left hand-written
#284 moved `panel_force_eqs` onto VortexStepMethod's panel kernel, so the frame SymAWE compiles is the one VSM evaluates. Two places still build that frame by hand.

### `wagner_reference_frame` — `src/aero_modes/common.jl:585-593`

Its docstring says it is "built the way `panel_force_eqs` builds a panel's axes, down to the `panel_span_signs` orientation, so the wing angle and the panel angles it shifts have the same sign". After #284 nothing enforces that — `panel_force_eqs` reads `VortexStepMethod.panel_axes` and this loop restates it — and the two already disagree about the normal: the loop takes it from the leading-edge step (`z_vec = x_unit × (left.LE_point .- right.LE_point)`) where OpenSourceAWE/VortexStepMethod.jl#273 takes it from the quarter-chord step. #294 measured those 10.7° apart on the SK100's raked tip.

Replace the loop body with

```julia
axes = VortexStepMethod.panel_axes(left.LE_point, left.TE_point,
                                   right.LE_point, right.TE_point,
                                   chord_weight, orient[i])
x_ref .+= axes.x_airf
z_ref .+= axes.z_airf
```

and drop `chord_vec`, `x_unit` and `z_vec`. `panel_axes` already applies `orient` to `z_airf` and not to `x_airf`, which is what the loop does by hand.

**To decide while doing it:** the loop hardcodes the 0.5 midpoint chord, where the panels pass their stored `chord_weight`. Passing the stored weight is the faithful match, and is a small behaviour change on a wing whose weights are not 0.5.

### `chord_frame_coordinates` — `src/aero_modes/live_polar.jl:70-86`

Same cross product (`normal = cross(x_axis, le_1 .- le_2)`), but a different case, and it may be right as it stands: it builds from `panel.corner_points` on purpose, so a replayed log frame restores it, and it sign-aligns to `panel.z_airf` with a fallback to it. The question is whether it should call `panel_axes` over those corner points rather than restate the cross product. Decide it rather than assume it; if the answer is yes, it is a second PR, not a second hunk.

### Not in scope

The quarter-chord normal itself is a VSM fix — OpenSourceAWE/VortexStepMethod.jl#273 — and reaches SymAWE with a VSM 5.x release and a compat widening past `VortexStepMethod = "4.3"`. #294 tried to patch SymAWE's copies instead and is closing in favour of that.

1-Bort-1:
### TL;DR

`wagner_reference_frame` built the wing's reference axes with its own cross product where `panel_force_eqs` reads `VortexStepMethod.panel_axes`, so nothing kept the frame the Wagner lag measures its one angle of attack in equal to the frame of the panels that lag shifts. It now calls the same kernel, at the same chord blend weight VSM gives each panel, and `panel_chord_width` next to it reads the kernel's `panel_chord` and `panel_span_vector` instead of restating those too.

### What was wrong

#284 moved `panel_force_eqs` onto VortexStepMethod's panel kernel and left this loop behind. Its docstring still claimed it was "built the way `panel_force_eqs` builds a panel's axes", and after #284 nothing enforced that — one function reads `VortexStepMethod.panel_axes`, the other restated it.

They had already drifted, in the input rather than in the formula. `panel_axes` leans a panel's chord direction towards its narrower neighbour by a blend weight, and VSM's own geometry pass builds every panel's stored axes with the weight `panel_chord_weight` gives it from the neighbouring panel widths (`wing_geometry.jl:205-215`). The loop hardcoded the 0.5 midpoint. On the 2plate kite with the billowing panel distribution the test fixture uses, the wing frame sat 0.0036° off the mean of the panels' own `x_airf` and 0.0043° off their mean `z_airf` — small, because that mesh is nearly uniform and barely swept, and unbounded in principle, because nothing tied the two together.

The normal is the other half of the issue and it is not touched here: in VortexStepMethod 5.1, the version `[compat]` pins, `panel_axes` still takes `z_airf` from the leading-edge step exactly as this loop did, so today the two agree on it. The 10.7° OpenSourceAWE/VortexStepMethod.jl#273 measures on a raked tip arrives when that fix is released and the compat widens — and it arrives in this frame on its own, which is the point of the change.

### What I changed

The loop body is `VortexStepMethod.panel_axes` at the panel's `panel_span_signs` orientation and its mesh blend weight; `chord_vec`, `x_unit` and `z_vec` are gone. The neighbour-width bookkeeping that turns widths into weights is now one function, `chord_blend_weights!`, so the build-time frame and the per-solve freeze in `store_chord_weights!` cannot disagree about a tip panel. That costs `store_chord_weights!` one n-element widths vector per VSM refresh, immediately after the nonlinear solve that refresh runs.

`panel_chord_width` came along, behaviour-preserving: its chord and width were the same kind of hand-written restatement of `VortexStepMethod.panel_chord` and `smooth_norm(panel_span_vector(...))`, in the very function the new code calls for the widths the weights need. Identical in Float64 — `smooth_norm` floors a norm at 1e-12 in quadrature, which is below an ulp of a panel width.

### The second frame the issue asks about

`chord_frame_coordinates` (`src/aero_modes/live_polar.jl`) should go through `panel_axes` too. It is #302 rather than a second hunk here, with the reading written up there. Two things decide it. Its `fraction` is compared at `src/aero_modes/pressure.jl:547` against contour-node fractions that `loft_contour_node` places with `panel.x_airf` and `panel.chord` — the `panel_axes` quantities — while `chord_frame_coordinates` measures with an unleaned mid-chord axis and `norm(chord_vec)`, so on a swept or tapered panel the two fraction scales being matched against each other are not the same scale. And `panel_axes` takes the corner points as arguments and never reads `panel.z_airf`, which `read_aero_log_points!` does not restore on a replayed log frame — the sign guard and the `z_airf` fallback are the one place a frame built from `corner_points` on purpose reads state the replay leaves stale. It has to pass `orient = 1`: VSM's `reinit!(::Panel)` bakes the wing-level span flip into the stored corner ordering, so applying the wing's sign again would double it.

### Where I would push back

Passing the mesh blend weight is a behaviour change on any wing whose panels differ in width, which is every non-uniformly panelled wing. I took it over bit-compatibility because the frozen mesh weight is not an approximation of what the panels use — it is the same number, from the same function, over the same widths — and because leaving one input of the frame at a literal is what let the two drift in the first place. A wing with the Wagner lag on will see its `deficiency` change in the last few digits.

Found and not fixed: `LivePolarState.control_fraction` is written at `src/aero_modes/live_polar.jl:185` and read nowhere, which is #302's file, and `.github/workflows/CI.yml:38` carries `fail-fast: false`, which is #307 — left alone here, since it is a deliberate setting and not mine to flip inside this diff.

### Verification

- [x] Reproduced first: the new `@testset "Wagner: the reference frame is the panels' own"` fails against unchanged code — `Unsteady aero | 40 passed, 2 failed`, `x_ref` 0.0036° and `z_ref` 0.0043° off the mean of the panels' own axes
- [x] `test/test_unsteady_aero.jl` red before, green after: 42/42 (5m43s), the frame testset 3/3 (juliaserver, exit 0)
- [x] `test/test_aero_modes.jl` 809/809 (3m27s) · `test/test_backend_parity.jl` 3/3 (2m32s)
- [x] All three re-run on the current head, against VortexStepMethod 5.1.0 under Julia 1.12.7, with every model in them rebuilt from scratch — the merge below moved the cache key to `v0.17.0_jl1.12`
- [x] Docs build clean — `docs/make.jl` runs through, the only warnings are the page and search-index size thresholds of #301
- [x] Up to date with `main` at fa29f14c. `reuse` and `jetls` are not installed on this box, so both lints are CI's
- [x] Local full suite on the merged head: PASS in 39 min, Julia 1.12.7, one cell of the matrix
- [x] GitHub CI on ef22f177: PASS, all seven jobs — the five test cells (1.12 on Linux, macOS and Windows, the kernel backend, and 1.13 on Linux, which replaced 1.11 in the merge below), Documentation and reuse-lint
- Benchmark: n/a — build-time geometry, evaluated once per model
- Risk: the blend weight is read off `wing.vsm_wing.refined_sections` where the panels read it off the live `BodyAerodynamics`; on a wing whose mesh is refined differently from its frozen sections the two would be different widths, and the new test is what catches it.

### Scope

+61 / -30 across four files, of which `src/aero_modes/common.jl` is the whole change: three functions in one region and no call site touched. The rest of the branch is two merges of `main`, at v0.17.0 and at the Julia 1.12/1.13 matrix move of #322; `CHANGELOG.md` was the only conflict in either, both times because both sides had written under the same heading, and `git diff main -- CHANGELOG.md` is the one `wagner_reference_frame` bullet. Follow-ups, neither stacked on this: #302 (`chord_frame_coordinates`, the second frame the issue asks about, open as #308), #307 (CI's `fail-fast: false`).

Closes #300 · task `SymbolicAWEModels.jl-300`


1-Bort-1 (commented):

**Good**

- The duplicate chord-blend boundary logic is extracted into `chord_blend_weights!`, which both `wagner_reference_frame` and `store_chord_weights!` now call; established by reading `src/aero_modes/common.jl` lines 493-501 and 1000-1007.
- `panel_chord_width` and `wagner_reference_frame` now route through `VortexStepMethod.panel_chord`, `panel_span_vector`, and `panel_axes` instead of restating the geometry; established by comparing the removed and added code in `src/aero_modes/common.jl` lines 472-610.
- New test `Wagner: the reference frame is the panels' own` asserts the wing frame equals the mean of the panels' own `x_airf`/`z_airf`; the PR card says it fails before and passes after, and the assertion matches the stated drift angles.
- Docstring updates, CHANGELOG entry, and `docs/src/private_functions.md` autodocs entry for `chord_blend_weights!` are all present in the diff.

1-Bort-1:
**Local full suite: PASS** (37 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
Heads up from #308 (issue #302), which is open off `main` alongside this one: both branches add the same equation to `src/aero_modes/common.jl` under names one `!` apart.

Here: `chord_blend_weights!(chord_weight, width)`, in place, neighbours at panel index ±1.
There: `chord_blend_weights(width, step)`, allocating, neighbours at ±`step` — `step = 1` is exactly this function, and `step = -1` is the case #302 needs, where the panel corners run along the wing's `spanwise_direction` against the panel index and the weight is the complement (`panel_chord_weight(wp, wo, wn) = 1 - panel_chord_weight(wn, wo, wp)`).

So whichever of us merges second should end up with one function, not two: keep the `step` argument, drop the in-place copy, and have both callers write `chord_weight .= chord_blend_weights(width, 1)`. That also hands `store_chord_weights!` back the zero-allocation refresh #308 costs it.

Neither branch can stack on the other at this point — both are pushed, and `agent stack` only moves a branch the remote does not already have. Flagging it here rather than leaving it to the merge.

1-Bart-1 (approved):

1-Bort-1:

PR #308, opened 2026-09-09
Build the live-polar chord frame with panel_axes over the panel corners, and refuse a wing whose sections run -y to +y

### LINKED ISSUE #302: cleanup: chord_frame_coordinates should build its frame with panel_axe

1-Bort-1:
### TL;DR

`chord_frame_coordinates` built its own chord frame by hand, and it was not the frame its result is compared against. It is now `VortexStepMethod.panel_axes` over the panel's corners with `orient = 1`. That is only right when a wing's sections run +y to -y, so, as Bart decided on this thread, any other order now errors.

### What was wrong

The fraction `chord_frame_coordinates` returns is matched against contour-node fractions in `build_station_point_map!`, and `loft_contour_node` places those nodes along `panel.x_airf` over `panel.chord`, which are the `panel_axes` quantities. The hand-rolled frame differed from that frame in three ways. It measured along the unleaned mid-chord axis, it divided by `norm(chord_vec)`, and it took its normal's sign from `panel.z_airf`. That third one reads exactly the state a replay leaves stale, because `read_aero_log_points!` restores the corners and nothing else.

On a panel with a 2.45° chord lean (blend weight 0.62) and 1.7% taper between its section chords, sections running +y to -y, a point at 40% chord and 3% of chord off the chord line reads as **(0.4051, 0.04789)** through the old frame. The offset is the camber increment a live polar is regenerated from, and it is out by a factor of 1.6: 0.4·sin(2.45°) = 0.017 chords of the along-chord displacement bleeds into the normal component, more than half the 0.03 being measured. The new frame reads (0.4, 0.03) exactly.

Give that panel a `z_airf` pointing the other way, which is what a replay can leave, and the old answer's offset flips sign. Through the real replay path on the fixture, the same flip moves every control-point offset from +0.01587 to −0.01587.

### Span order

With the sign guard gone, the frame's orientation comes from the order of the corners alone. `reinit!(::Panel)` swaps a panel's two sections when `wing_span_flip` is −1, but a replayed log writes the corners back in structural order. On a wing whose sections run -y to +y, the replayed frame therefore turns over (`|replayed − live| = 0.2` on a forced flip, against 5.3e-15 at flip = 1). An earlier round of this branch handled that with `spanwise_corners` and a `step = -1` walk in `chord_blend_weights!`.

Bart ruled on this thread that the only supported order is VortexStepMethod's, +y to -y, that anything else should error, and that the code handling the opposite order should go. So:

- `spanwise_corners` and `chord_blend_weights!`'s `step` are deleted. `chord_frame_coordinates(panel, chord_weight, pos_b)` passes `panel_corners` straight to `panel_axes` with `orient = 1`, as #302 described.
- `check_span_order(wing)` errors unless the first unrefined section's leading edge lies further along `spanwise_direction` than the last one's. VSM does not do this itself: its YAML loader and `refine!` sort or reverse a list quietly, and `reinit!` flips the panels. So the check runs at the three places a wing's sections are written: `transform_vsm_sections_to_body!`, `match_aero_sections_to_structure!` (before `refine!` can reverse the list) and `update_vsm_wing_from_structure!`.
- No fixture or example in the repo stores its sections in the other order: both `2plate_kite` aero geometries start at +y, and so do the structural YAMLs the tests use.

The rest of the opposite-order handling was `panel_span_signs` → `orient`, reaching the symbolic force equations, `AeroPanel`, the kernel backend's `:_down` key and `test_flow_curvature.jl`. #355 removed it for #353, stacked on this branch, and Bart merged it into this branch, so it arrives with this pull request as commit `3598857a` (+32 / −69, deletions only). Its reasoning and its own push-back are on #355.

### What else changed

The blend weight comes from `chord_blend_weights!` over `panel_span_width`, the span widths of the corners, so it survives a replay the way `panel.width` does not (`corner_chord_weights`). `store_chord_weights!` and `wagner_reference_frame` keep main's version unchanged.

`LivePolarState.control_fraction` is deleted. It was written and never read. The `@warn` that looked like a reader was measuring the station groups' own lengths, so it now measures them directly.

### The divisor does not wash out

#302 asked for evidence, not an assumption, that changing the divisor from `norm(chord_vec)` to `panel_chord` washes out through the reference differencing and `chord_residual!`. The test used the same swept, tapered panel: three control points on the chord line at 10/50/90%, the mid-chord one bumped 2% of chord off it, run through `control_point_deflection` and `chord_residual!` in three frames.

```
frame                                fractions                      max|residual|
new  (panel_axes, panel_chord)       [0.1,      0.5,      0.9]        0.01933449
mix  (panel_axes, norm(chord_vec))   [0.101703, 0.508516, 0.915328]   0.02010091
old  (mid-chord axis, same divisor)  [0.10161,  0.50892,  0.91449]    0.02010088
```

The lean does wash out of the deflection: old and mix differ by 1.5e-6 relative, because the reference and the deformed point share the frame. The divisor does not. It scales the ordinates and stretches the stations they sit on by the same 1.7%, which is not the affine part `chord_residual!` removes, and the camber residual moves by 3.8%. At the `build_station_point_map!` call site there is no differencing at all: the fraction is matched against `xc[k]` directly, and it was reading 0.1016 for a node at 0.1.

### Where I would push back

Nothing in this repo's fixtures moves. The 2plate kite's sections are parallel and equally spaced, so every blend weight is 0.5 and `norm(chord_vec) == panel_chord`. Across its 10 panels the frame moves by 2.2e-16 in fraction and 7.9e-31 in offset. The defect is real but latent, which is why a hand-built panel is the test that proves the change. For the same reason the replay testset is a coverage test: that fixture's control points sit on the chord line, so a replayed offset is 0 either way.

`check_span_order` compares only the first and last sections, as VSM's `wing_span_flip` does. A section list that doubles back in y between the tips passes it. `panel_span_signs` tests each panel on its own, and #353 is where that difference gets decided.

### Verification

- [x] Reproduced first: the old frame reads the 40%-chord, 3%-offset probe on the test's swept, tapered panel as `(0.4051346625509663, 0.04788860940091752)`, and `(0.4051346625509663, -0.04788860940091752)` with `z_airf` reversed; the new one reads `(0.4, 0.030000000000000002)` both times. Through the replay path on the fixture a stale `z_airf` moves every control-point offset `+0.015868246442849` → `-0.015868246442850`
- [x] `test/test_live_polar.jl` red before, green after: the new assertions run against a verbatim copy of the old function, which printed the numbers above; 158 pass after (the reversed-corner assertion went with `spanwise_corners`)
- [x] `test/test_match_aero_sections.jl` red before, green after: with `check_span_order` stubbed out, "a wing whose sections run from -y to +y is refused" fails 3 of 3 (`No exception thrown` from `update_vsm_wing_from_structure!`, `match_aero_sections_to_structure!` and `transform_vsm_sections_to_body!`); with it, 77 + 70 pass
- [x] On `535053b9` (VSM 5.1.0, Julia 1.12.7, juliaserver, exit 0, fresh model cache): `test_live_polar` 158 · `test_match_aero_sections` 77 + 70 · `test_pressure_aero` 80 + 20 · `test_continuous_modes` 92 · `test_continuous_aero` 52 · `test_flow_curvature` 76 · `test_unsteady_aero` 42 · `test_aero_modes` 809 · `test_backend_parity` 3; `test_transform` 357 on the merge before #355
- [x] Docs built locally on `53f28e95`: no `missing_docs`; `private_functions.html` is 510.65 KiB against the 600 KiB `size_threshold`, with only the size warnings that predate this branch (#301)
- [x] Up to date with main as of `535053b9`: merged the v0.18.1 release, #338, #340, #342 and #351, and #355 from this branch's remote. Conflicts in `CHANGELOG.md` only: the release renamed `Unreleased`, so this branch's two bullets sit in a new `## Unreleased` → `### Fixed`, above main's `### Changed`
- [x] Local full suite on `535053b9`: **PASS** (62 min, Julia 1.12.7, one matrix cell, on a box running four other suites)
- [ ] GitHub CI on `535053b9`, run 35328044022: six of seven green (Documentation, reuse-lint, the three Julia 1.12 monolith jobs on ubuntu, macOS and Windows, Julia 1.13 ubuntu monolith); the 1.12 ubuntu kernel job still running. The previous head `6686fcce` was green on all seven (run 35130007400)
- Benchmark: n/a. `check_span_order` is one 3-vector subtraction and dot product per section write; `corner_chord_weights` is two `n_panels` vectors per `update_live_deflection!`, against a NeuralFoil pass in the same call
- Risk: a wing built outside this repo's fixtures with sections running -y to +y now errors at setup instead of running on an order that was never supported

### Scope

+203 / −116 across 15 files. `src/aero_modes/{common,live_polar,pressure}.jl` is the chord frame; `src/vsm_refine.jl` and the one line in `transform_vsm_sections_to_body!` are the order check. Tests: two new testsets in `test_live_polar.jl`, one in `test_match_aero_sections.jl`. The other six files, and the rest of `common.jl`, `pressure.jl` and `continuous.jl`, are #355's `3598857a`.

Closes #302 · task `SymbolicAWEModels.jl-302`


1-Bort-1:
**CI: `Julia 1.12 - ubuntu-latest - x64 - kernel - pull_request` failed** on `ffc29be90`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34346049698/job/102447777526

1-Bort-1:
**Local full suite: FAIL** (18 min, Julia 1.13.0, one cell of the matrix)

```
  flap + continuous pressure aero   |   18           18     40.8s
  point flap drives the polars      |   10           10   1m13.1s
  Flap δ + beam-anchored point      |   28           28     15.4s
  Flow curvature                    |   76           76     32.2s
  Getter allocations                |    6            6     12.7s
  Getter allocations with a wing    |    4            4     29.3s
  Heading - Circular Path           |   72           72      0.2s
  Heading - Horizontal Circle       |   48           48      0.1s
  Heading - Special Cases           |   22           22      0.2s
  Heading - Near-Vertical Placement |   10           10      0.6s
  Testing helper functions...       |    6     1      7      1.2s
RNG of the outermost testset: Random.Xoshiro(0x5e1c19051b144d0f, 0x1695904cf9fd14d5, 0xfa4469ccd6659392, 0xe85627e37a365f95, 0xdf8df22490840a7b)
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Types.jl:68
    @ Pkg.Operations ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Operations.jl:3148
  [3] test
    @ ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Operations.jl:3026 [inlined]
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:586
  [5] kwcall(::@NamedTuple{io::IOContext{IO}}, ::typeof(Pkg.API.test), ctx::Pkg.Types.Context, pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:562
  [6] test(pkgs::Vector{PackageSpec}; io::IOContext{IO}, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:172
  [7] test(pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:161
  [8] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::PackageMode, subdir::Nothing, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:189
  [9] test()
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:178
 [10] top-level scope
    @ none:1
 [11] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [12] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:310
 [13] _start()
    @ Base ./client.jl:577
```

1-Bart-1:
We had problems with signs / spanwise ordering before, so now we decided that the only supported ordering is -y --> +y. This PR won't change the behaviour of such a wing right?

1-Bort-1:
Right on signs and spanwise ordering; not right on the chord scale, which is the whole point of the change. Measured on the branch head `e5caca93`, 2plate_kite fixture, `spanwise_direction = [0,1,0]`, `wing_span_flip = 1`:

    panels spanwise_corners reorders              0 of 10
    old dot(z_axis, panel.z_airf) < 0 guard       fired 0 times
    |corner_chord_weights - store_chord_weights!| 0.0  (every weight 0.5)
    max |old fraction - new fraction|             2.2e-16
    max |old offset   - new offset|               7.9e-31

So `spanwise_corners` hands `panel_axes` exactly the stored corner order, live and replayed alike, and dropping the old `z_airf` sign guard flips no normal.

One correction to the premise, worth knowing: the order the panels actually end up in is **descending** y, +y -> -y, whatever the geometry file says. `refine!` sorts `by=span_order_key, rev=true` when `sort_sections=true`, and on the structural path (`sort_sections=false, recompute_mapping=true`, src/vsm_refine.jl:229) it calls VSM's `normalize_span_order!`, which reverses an ascending list. `wing_span_flip` is `dot(first.LE - last.LE, spanwise) < 0 ? -1 : 1`, so that order gives +1 and `reinit!(::Panel)` never swaps sections. I found no path in this repo that reaches -1; forcing it needs a negated `spanwise_direction`, which VSM 5's loader rejects.

What does change is a **swept or tapered** wing, in either ordering, and that is #302's defect. Synthetic panel, sections in descending y so nothing is reordered, probe placed at 40% chord and 3% offset by construction:

    panel_chord 0.918501   norm(chord_vec) 0.903120   (1.7% apart)
    x_airf leans 2.61 deg off the mid-chord axis
    old -> (0.40778, 0.01198)
    new -> (0.40000, 0.03000)

The fixture cannot show that: its sections are parallel and equally spaced, so every blend weight is 0.5 and `panel_chord == norm(chord_vec)`.

Your rule does make one thing in the diff look like dead weight. If flip = -1 is unreachable, then `spanwise_corners` and the `step = -1` branch of `corner_chord_weights` are machinery for a wing the repo cannot build, and #302 taken literally — `panel_corners` straight into `panel_axes` with `orient = 1` — is a smaller diff that behaves identically on every supported wing. I kept them because without them a section-swapped wing's frame turns over on replay (`|replayed - live| = 0.2` on the forced-flip probe). Say which you want and I will push it; I am not touching it on my own guess.

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `63cf6b72d`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34938527751/job/104281644746

1-Bart-1:
Fix the conflicts

1-Bort-1:
Done in 04df8268 and ea0882ed; #308 is MERGEABLE and all seven CI jobs are green on the head.

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `ea0882ed7`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34955360320/job/104335985649

1-Bart-1:
Yes, we should assume the spanwise ordering is the one used in vsm (+y to -y right). And error if it is not this order. And remove code trying to deal with opposite ordering.

CHANGELOG.md
```diff
@@ -15,6 +15,19 @@
...
+### Fixed
+- A live polar's control points are measured in the panel's own `panel_axes`
+  frame over its corners, which is the frame its contour nodes are lofted along.
+  The hand-built frame they used before measured along the unleaned mid-chord
+  axis over `norm(chord_vec)`, so on a swept or tapered panel the chord fraction
+  and the camber offset were on a different scale from the nodes they are
+  matched against.
+
+### Changed
+- BREAKING: a wing whose aerodynamic sections, or the structural stations they are
```
1-Bart-1: This is not breaking. It already did not work. Now it is just errorring as well when it is wrong.

1-Bort-1: Fixed in 6686fcce: the entry now sits under Fixed, without BREAKING.

1-Bort-1:
Done on #308: cc1b0714 removes spanwise_corners and the step = -1 walk, and 53f28e95 adds check_span_order, which errors on -y to +y sections. The remaining opposite-order code (panel_span_signs / orient / the :_down kernel) is #353, stacked on #308.

1-Bart-1 (approved):

Issue #305, opened 2026-09-09
test_helpers' manifest-staleness check is an mtime comparison, so a dev'd dependency turns the local suite red


1-Bort-1:
`test/test_helpers.jl:50-51` asserts that both tracked `Manifest-v1.1x.toml.default` files are no older than `Project.toml`, as a "you changed the project, regenerate the manifests" reminder (#199). The proxy is the file's mtime, and mtime moves whenever `Pkg` rewrites `Project.toml` — even when it writes back the identical bytes.

`Pkg.develop`/`Pkg.free` of a locally checked-out dependency does exactly that, and that is the prescribed workflow for cross-repo work (`AGENTS.md` §4, "Work in a dependency"). The result is a worktree whose `Project.toml` is byte-identical to `main` and whose full suite is permanently red, with no way back that an agent is allowed to take: regenerating the `.default` manifests is `./bin/update_default_manifests`, which is a PR of its own.

Seen on #291: `mtime(Manifest-v1.12.toml.default) = 1.7888818752933135e9` against `project_mtime = 1.788948339274766e9` — 18 hours apart, while `git diff origin/main -- Project.toml Manifest-*.default` is empty. CI never sees it, because a fresh checkout stamps every file within the same second.

What I would do instead: check the content the reminder is really about — every direct dependency in `Project.toml` appears in each `.default` manifest at a version its `[compat]` entry allows. That catches the case the mtime check was written for (a dep added or a bound tightened without re-resolving), catches it in CI as well as locally, and does not fire on a file that was rewritten with the same bytes.

Found while working #291; not fixed there because it is unrelated to that change.

1-Bort-1:
The same helper has a second blind spot, and it cost a full local suite run on #309 today.

`test/test_helpers.jl:50-51` checks the tracked `.default` manifests against `Project.toml`. Nothing checks the *live* `Manifest-v1.1x.toml` against the `.default` it was copied from, so a worktree can sit a major version behind its own code and only the suite finds out.

[...]

PR #309, opened 2026-09-09
cleanup: bin/install has no -y, so the box cannot install a worktree



### LINKED ISSUE #303: cleanup: bin/install has no -y, so the box cannot install a worktree
cleanup: bin/install has no -y, so the box cannot install a worktree

`./bin/install -y` exits 1 with `Invalid parameter!` and the usage text, so every worktree cut for this repo comes up on an environment that was never installed and `using SymbolicAWEModels` fails until an agent installs it by hand.

rules/AGENTS.md §2 describes the flag: `-y` "runs it with no terminal — no menu, no `read`, the Julia already on the machine — and without it the box waits on a prompt nothing will answer". The script's only no-prompt path today is `+1.11` / `+1.12` / `--version X.Y`, which names a version rather than taking the one already active.

The same section says none of the three scripts changes anything outside the repo. Two places in this one do:

- it appends `alias jl='bin/run_julia'` to `~/.bashrc`, or to `~/.zshrc` / `~/.bash_profile` on macOS;
- `juliaup add "${_desired_julia}"` installs a Julia channel.

So `-y` should select the active Julia, skip the menu and the `read`, and leave both of those alone.

1-Bort-1:
### TL;DR

`./bin/install -y` exited 1 on the argument loop, so every worktree cut for this repo came up on an environment that had never been instantiated and `using SymbolicAWEModels` failed until somebody installed it by hand. The script now takes the `-y` that rules/AGENTS.md §2 describes: the Julia already active, no menu and no `read`, and neither of the two writes outside the repository.

### What was wrong

`-y` was never a case in the option loop, so it fell through to `*)` and printed `Invalid parameter!` with the usage text. The only no-prompt path the script had was `+1.11` / `+1.12` / `--version X.Y`, which names a version rather than taking the one already on the machine, and even that path still ran `juliaup add` and appended `alias jl='bin/run_julia'` to the shell startup file — so there was no way to run this installer that left the machine as it found it.

### What `-y` does

With no `+X.Y` or `--version` it takes the major of `julia --version` and skips the menu, so the box's Julia is the one it installs for. `juliaup add` and the alias append run only on the interactive path. Everything from the version choice onwards is untouched: `-y` differs from a menu run only in what it does not do.

The interactive run is as it was — same menu, same default, same channel install, same alias.

### Rode along

Two behaviour-preserving cleanups in the same file, both in code the flag had to touch anyway. `_version_specified` and the `_current_julia` / `_current_major` else branch were assigned and never read; they are gone, and the version-detection expression from that dead branch is what `-y` now uses. The alias block was one append written twice, once for Linux and once for macOS, differing only in which startup file it picks; it is now one file choice and one append.

### Where I would push back

§2's sentence reads unconditionally — "None of the three changes anything outside the repo — no `juliaup default`, no line appended to `~/.bashrc`" — which would delete the alias block outright rather than gate it, and `VortexStepMethod.jl`'s copy of this script (the same lineage: same flags, same `_yes_to_all`) carries no such block at all. #303 asks only that `-y` leave it alone, so that is what this does. Say the word and it goes.

Second thing worth knowing: nothing in CI or `test/` runs `bin/install`, which is why the drift sat here unnoticed — `setup-test.yml` covers the end-user path (`test/setup_integration.jl`) only. A step running `./bin/install -y --skip-subprojects --no-precompile` costs a second and a half and would catch the next one.

Third, found while this branch was open: the local suite went red on the merge commit and it was the environment, not the diff. Merging `main` in brought VortexStepMethod 5 (`5a89aa18`) and a regenerated `.default` manifest, but nothing re-installs a worktree after a merge — so it kept running the 4.3.1 it had been installed with, and `test_live_polar.jl` died 17 minutes in on a `FieldError` saying `WingSettings` has no field `airfoil`. `test/test_helpers.jl:50-51` compares the `.default` manifests' mtime against `Project.toml` and nothing compares the *live* manifest against the `.default` it is copied from, so a worktree can sit a major version behind its own code with no cheap signal. That check is #305's subject and I have put the case there rather than open a second issue.

### Verification

- [x] Reproduced first: `./bin/install -y` → `Invalid parameter!` + usage, exit 1
- [x] Green after: `./bin/install -y < /dev/null` ran to `Installation complete!`, exit 0, choosing `julia +1.12` (the active channel) with no menu and no `juliaup add`; `using SymbolicAWEModels` then loads 0.16.0 in this worktree's session. Run twice — the second time on the merge head `a05b9f2e`, where it is what took the worktree off VortexStepMethod 4.3.1 and onto the 5.0.0 the tracked manifest pins, which is the flag doing the job the issue asks of it
- [x] Nothing outside the repo moved: `~/.bashrc` md5 `4b4b037a…` before and after, `juliaup status` still `1.11` / `*1.12`
- [x] `./bin/install -y --skip-subprojects --no-precompile` exits 0 in 1.3 s (the resolve-only path)
- [x] `./bin/install -h` lists `-y`; an unknown flag still exits 1; `bash -n bin/install` clean
- [x] Local full suite: PASS twice, Julia 1.12.7, one matrix cell — 28 min on the pre-merge head `54e3df42`, 32 min on the merge head once the worktree had been re-installed. The red run between the two is the third note above: `test/test_live_polar.jl` against a worktree still holding VortexStepMethod 4.3.1, which errors with or without this diff and passes 153 / 153 in 43.9 s the moment the manifest matches the code
- [x] GitHub CI on the merge commit `a05b9f2e`: all seven checks green — `Documentation`, `reuse-lint`, and the five-cell matrix (1.11 ubuntu, 1.12 ubuntu monolith and kernel, macOS aarch64, Windows)
- [x] Up to date with `main`: merged `ad49927a` in; the only conflict was that #288 and this branch had each added the first bullet under Unreleased / Added, and both bullets are kept
- Docs build: not run locally — the diff is one bash script plus one changelog entry, no Julia symbol changed, and no docs page lists the install flags. The Documentation job on GitHub CI is green.
- Risk: `-y` with an explicit `+1.11` no longer installs a missing channel, so on a machine that lacks it the run stops at the first `julia +1.11` instead of fetching it. The box's own call names no version and takes the active one.

### Scope

+46 / -34 across 2 files: `bin/install`, and one entry under Unreleased / Added in `CHANGELOG.md`. No stack — no open PR touches `bin/`.

Closes #303 · task `SymbolicAWEModels.jl-303`


bin/install
```diff
@@ -134,15 +149,14 @@ if [[ -z "$_desired_julia" ]]; then
...
-else
-    # Version was specified via command line
-    _current_julia=$(julia --version 2>/dev/null | grep -oE '1\.[0-9]+' | head -1)
-    _current_major="${_current_julia%.*}"
 fi
 
-echo "Ensuring Julia ${_desired_julia} channel is installed..."
-juliaup add "${_desired_julia}"
-echo
+# Installing a channel writes outside the repository, so -y never does it.
```
1-Bort-1, line 154: MINOR: The comment explains why the guard exists (juliaup writes outside the repo), which violates the 'never write why' rule in §3/§5; this rationale already belongs in the PR description.

bin/install
```diff
@@ -283,34 +297,29 @@ if [ -f ".JETLSConfig.toml" ] && [ -f ".JETLSConfig.toml.default" ]; then
...
-if [[ "$(uname -s)" == "Linux" && -f ~/.bashrc ]]; then
-    if ! grep -q "alias jl='bin/run_julia'" ~/.bashrc; then
-        echo "alias jl='bin/run_julia'" >> ~/.bashrc
-        echo
-        echo "Added alias 'jl' to ~/.bashrc"
-        echo "Run 'source ~/.bashrc' or restart your terminal to use it"
-    fi
-elif [[ "$(uname -s)" == "Darwin" ]]; then
-    # On macOS, check for zsh (default since Catalina) or bash configuration files
+# Add the alias to the shell configuration file, which lives outside the
```
1-Bort-1, line 300: MINOR: The comment explains why the alias block is gated (it writes outside the repo), which violates the 'never write why' rule in §3/§5; the PR description already states this intent.

1-Bort-1 (commented):

**Good**

- New -y/--yes option parses, updates usage/help, and runs the active Julia channel with no prompt; PR card shows the before failure and after green path.
- Dead assignments _version_specified, _current_julia, and _current_major are removed, and the version-detection expression is reused for the -y path.
- The shell alias block is collapsed from two near-identical OS copies into one file-selection path and one append, removing duplication.
- CHANGELOG records the new flag under Unreleased/Added with the no-menu/no-outside-changes behavior.
- No tracked generated files, manifest edits, version bump, or dependency were added.

**Not good**

- `bin/install:154` — The comment explains why the guard exists (juliaup writes outside the repo), which violates the 'never write why' rule in §3/§5; this rationale already belongs in the PR description.
- `bin/install:300` — The comment explains why the alias block is gated (it writes outside the repo), which violates the 'never write why' rule in §3/§5; the PR description already states this intent.
- -y combined with +X.Y/--version skips juliaup add even if the channel is missing, which the PR card notes as a risk rather than intended behavior.
- bin/install has no CI or automated test coverage, so the parse failure went unnoticed; the PR card already proposes a one-second CI step.
- The comment on line 104 restates the following if block rather than stating a non-obvious fact.

1-Bort-1:
**CI: `Julia 1.11 - ubuntu-latest - x64 - monolith - pull_request` failed** on `54e3df425`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34346078324/job/102447871932

1-Bort-1:
**Local full suite: PASS** (28 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

1-Bort-1:

1-Bart-1 (approved):

PR #315, opened 2026-09-09
Nothing tests that a transform lands on its heading near the vertical


### LINKED ISSUE #312: Nothing tests that a transform lands on its heading near the vertical

1-Bort-1:
### TL;DR

Adds a testset to `test/test_heading_calculation.jl` that places a body 51 m up in CAD, a hair off the transform's rotation axis, runs `reinit!` and asserts the body lands on `transform.heading`. #254's 180° flip no longer reproduces, but nothing guarded the fix, and the frame degeneracy it rode on is still live.

### What was unguarded

`OpenSourceAWE/SymbolicAWEModels.jl@510a4232` ended #254 in two places: `apply_azimuth_elevation!` now rotates the current radial onto the target one with `min_rotation` instead of composing two tether frames, and `apply_heading!` measures the body's own `Q_b_to_w` instead of re-applying the azimuth/elevation rotation that had already been applied. Neither has a test that would notice it going back.

The rest of the heading coverage cannot notice: `test/test_heading_calculation.jl` is pure geometry — it builds no `SystemStructure` and never calls `reinit!` — and `test/test_transform.jl` round-trips `transform.heading` only at elevation 80° with the wing well off the axis.

The degeneracy those two changes route around is unchanged and still reachable. At 51 m standoff `calc_R_t_to_w` (`src/generate_system/helpers.jl:210`) gives tangent frames exactly 180° apart for a millimetre of horizontal offset either way:

```
calc_R_t_to_w([+1e-3, 0, 51])[:, 2] = [-0.0,  1.0, 0.0]
calc_R_t_to_w([-1e-3, 0, 51])[:, 2] = [-0.0, -1.0, 0.0]   → 180.0°
```

So the new testset is what stands between that frame and the placement path.

### The test

One `Body` at `base_pos + [dx, 0, 51]` in CAD, a co-located `BODY_STATIC` point as the transform's `rot_point`, a `Transform` at 70° elevation / 20° azimuth / 30° heading, `reinit!`, then `calc_heading` on the placed body's world orientation — asserted equal to `transform.heading` for `dx` of `0`, `±1 mm` and `±1 cm`, from both an origin and a non-origin base. It builds no `SymbolicAWEModel`, so it compiles nothing and adds no model-cache entry; the whole testset runs in 0.1 s.

The construction is `test/test_rigid_body.jl:148`'s idiom, the one place in the suite that already puts a plain `Body` through a `Transform` at the `SystemStructure` level. `Settings` has to come through the process-wide data path, so the file reads `data/2plate_kite` and puts the path back as it found it rather than writing a second copy of a settings YAML — the suite has two of those already.

### Against #314

#314 rewrites the `apply_heading!` this test pins, so I ran the new testset with `src/system_structure/transforms.jl` taken from that branch: `10/10` pass, file total `152/152`. The two do not touch the same file and are not stacked; whichever lands first, this testset is the thing that would have caught #314 changing behaviour, and it says it does not.

### Where I would push back

The assertion is self-consistent by construction: `apply_heading!` measures the heading with the same `calc_R_t_to_w` the test measures it back with, at the *target* position where the frame is well-defined. That is what makes it a placement test and not a frame test — it pins the contract (`reinit!` puts the body on `transform.heading`) and is blind to the frame convention itself. A test for the degeneracy proper belongs beside `calc_R_t_to_w`, and #313 is already in that file.

### Verification

- [x] Reproduced first: n/a — the bug does not reproduce on `main`; the frame degeneracy it rode does, quoted above.
- [x] Red before green: with `apply_heading!`'s no-ref-points branch reverted to its pre-`510a4232` form, `0 passed, 10 failed` — `0.852` rad for `dx ≥ 0` against a target of `0.524`, and `-2.289` rad for `dx < 0`, i.e. the 18.8° double-application error *and* the 180.0° flip. Restored, `10/10` pass.
- [x] `test/test_heading_calculation.jl`: `152/152` pass, 0.5 s (juliaserver, exit 0). Re-run in a fresh module namespace after the import trim: `152/152`.
- [x] With #314's `transforms.jl`: `152/152`, near-vertical `10/10`.
- [x] Data path is left as found: verified from a `mktempdir` ambient path, `before == after`.
- [x] Up to date with `origin/main` (0 behind). Lines ≤ 92 chars.
- [ ] Local full suite: not run — the diff touches no `src/`. `agent ci-local` is running; last turn's attempt was reaped by the box's OOM killer before it produced output. GitHub CI triggers on this pull request.
- [ ] Docs: not built — no `src/` or `docs/` change and no new public symbol, so `make.jl` has nothing new to resolve. Changelog: no entry — test-only, nothing user-visible. REUSE lint: not run, `reuse` is not installed on the box; the file's SPDX header is untouched.
- Risk: the file now sets the process-wide data path to read `Settings` and restores it. A later test file that depended on the ambient path would break if that restore ever stopped happening.

### Scope

+54 / -1 in one file, `test/test_heading_calculation.jl`, all of it the new testset and the `Settings` bootstrap it needs. No `src/` change. Not stacked: #314 is open on `src/system_structure/transforms.jl`, a file this diff does not touch, and this branch was already pushed before it existed.

Closes #312 · task `SymbolicAWEModels.jl-312`

### DIFF
--- a/test/test_heading_calculation.jl
+++ b/test/test_heading_calculation.jl
@@ -30,8 +30,7 @@ using LinearAlgebra
 using KiteUtils
 using SymbolicAWEModels
 using SymbolicAWEModels: smooth_normalize, sym_calc_R_t_to_w,
-    calc_heading, calc_R_t_to_w, Point, Body, Transform,
-    SystemStructure, reinit!, quaternion_to_rotation_matrix
+    calc_heading, calc_R_t_to_w, reinit!, quaternion_to_rotation_matrix
 
 # `Settings` reads through the process-wide data path; leave it as found.
 prev_data_path = get_data_path()
--- a/test/test_heading_calculation.jl
+++ b/test/test_heading_calculation.jl
@@ -8,6 +8,10 @@ Verifies that `calc_heading` correctly projects the body x-axis
 onto the tangent plane of the tether sphere and returns the
 angle from the elevation direction (x_t) toward azimuthal (y_t).
 
+Also places a body through `reinit!` from a near-vertical CAD
+pose and verifies it lands on the transform's heading, where the
+tether frame the pose sits in is degenerate.
+
 Tests run with both origin and non-origin base positions to
 verify heading is computed relative to the sphere center.
 
@@ -23,8 +27,17 @@ end
 
 using Test
 using LinearAlgebra
+using KiteUtils
+using SymbolicAWEModels
 using SymbolicAWEModels: smooth_normalize, sym_calc_R_t_to_w,
-    calc_heading, calc_R_t_to_w
+    calc_heading, calc_R_t_to_w, Point, Body, Transform,
+    SystemStructure, reinit!, quaternion_to_rotation_matrix
+
+# `Settings` reads through the process-wide data path; leave it as found.
+prev_data_path = get_data_path()
+set_data_path(joinpath(dirname(@__DIR__), "data", "2plate_kite"))
+set = Settings("system.yaml")
+set_data_path(prev_data_path)
 
 # Support selective test execution via command-line args
 const test_patterns = isempty(ARGS) ? String[] : ARGS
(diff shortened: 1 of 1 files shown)


test/test_heading_calculation.jl
```diff
@@ -23,8 +27,16 @@ end
...
 using LinearAlgebra
+using KiteUtils
+using SymbolicAWEModels
 using SymbolicAWEModels: smooth_normalize, sym_calc_R_t_to_w,
-    calc_heading, calc_R_t_to_w
+    calc_heading, calc_R_t_to_w, reinit!, quaternion_to_rotation_matrix
+
+# `Settings` reads through the process-wide data path; leave it as found.
+prev_data_path = get_data_path()
+set_data_path(joinpath(dirname(@__DIR__), "data", "2plate_kite"))
```
1-Bort-1, line 37: MINOR: `set_data_path`/`get_data_path` restore the process-wide path only on the happy path; an exception before line 39 leaves the suite-wide data path changed and could break later tests. Wrap the `Settings` read in `try ... finally set_data_path(prev_data_path) end`.

1-Bort-1 (commented):

**Good**

- Adds regression coverage for the #254 near-vertical heading placement bug, which the existing geometry-only tests cannot catch.
- Verified red-before-green against the pre-fix `apply_heading!` and passes against both `main` and #314's transforms branch.
- Single-file, test-only change with no `src/` or public API churn, reusing the established `SystemStructure` construction idiom.
- Integrates with the file's selective execution pattern via `should_run_test("vertical")`.

**Not good**

- `test/test_heading_calculation.jl:37` — `set_data_path`/`get_data_path` restore the process-wide path only on the happy path; an exception before line 39 leaves the suite-wide data path changed and could break later tests. Wrap the `Settings` read in `try ... finally set_data_path(prev_data_path) end`.

1-Bort-1:
**Local full suite: PASS** (41 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

Issue #316, opened 2026-09-09
test_twist_alignment asserts on a state reached by driving the VSM to the edge of convergence


1-Bort-1:
`test/test_twist_alignment.jl:112-123` ramps steering for 60 steps and then checks chord alignment on whatever state that lands in. The comment above the loop already says what that state is: "Holding this steering runs the station twist away until the VSM solve diverges, so a settled operating point does not exist for this 2-plate config." The loop stops next to that divergence, and which side of it a run lands on is decided by the machine rather than by the code.

The alignment assertions are right and tight (`angle < 1e-5` deg, `length_err < 1e-10`) and they hold wherever a run gets that far. What is wrong is where the ramp stops.

It has gone red twice, both times only in `Julia 1.12 - ubuntu-latest - x64 - monolith`, on two unrelated branches:

- `feat/point-flap-delta` (#288), `93ac34cd`, 2026-09-08, [job 102146513513](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34250901786/job/102146513513) — `AssertionError: VSM solve failed (non-converged or non-finite) on wing 1 [eltype=Float64]` at `test_twist_alignment.jl:123`. The other five matrix cells of that run failed on `test_live_polar.jl:38` and nowhere else. `93ac34cd` is an ancestor of `main` (merged as `ad49927a`); the branch head after it and the merge commit both passed the same job.
- `agent/313-…` (#314), `c3b2b562`, 2026-09-09, [job 102647056275](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34405402473/job/102647056275) — the same assertion at the same line. #314 deletes dead code only; running the test with the parent commit's `src/system_structure/transforms.jl` swapped back in reproduces the same numbers to the last bit.

The same line has also taken the model out twice on `feat/live-polars` a different way — `AssertionError: Solver unstable at t=3.2016: DtNaN` (`2bedcba2` [job 101872595398](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34164445079/job/101872595398), `89b22128` [job 101841764152](https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34153952802/job/101841764152)). Three branches, two failure modes, one line.

### Why it is the ramp and not the code

Six runs of the test, five of them on `c3b2b562`:

| run | static twist L | dynamic twist L / C / R |
[...]

PR #317, opened 2026-09-10
Make a point's wing reference mean one thing

### LINKED ISSUE #238: Code crashes if wing_idx of a point is zero
According to Bart, points do not need to belong to a wing. But if they don't, the code crashes.

1-Bort-1:
### TL;DR

A point's `wing_idx` meant two things at once — the wing the point belongs to, and a default of 1 that a point which named no wing got anyway — so the monolith could not ask a point which wing it was in and asked the *model* whether it had any wing instead, crashing with a `BoundsError` on index 0 the moment a genuinely wingless point met a model that had a wing. `wing_idx` now means only the first of those, `0` is "no wing", and both whole-system guards and the "use the first wing" fallback behind every non-wing point's `va_b` are gone.

### What was wrong

`wing_idx == 0` was already the documented "no wing" sentinel: `Point`'s `wing_ref` field said so, `resolve_ref` mapped both `0` and a missing reference to `0`, and the `KernelBackend` honoured it in `rigid_wing_node` and `kinematic_wing_of`. But `Point`'s constructor handed every non-`BODY_STATIC` point that named no wing a `wing_ref` of `1`, so in practice almost nothing was ever `0` and the monolith's equation generator never had to handle it.

Reproduced on the 2plate particle kite with `kcu` and `ground` given `wing_idx: 0`:

```
ERROR: BoundsError: attempt to access ...BasicSymbolic... at index [:, 0]
  [6] point_eqs!(...)  @ src/generate_system/point_eqs.jl:323
  [7] create_sys!      @ src/generate_system/create_sys.jl:175
 [11] init!            @ src/model_management.jl:616
```

The free-particle branch decided whether to build the point's body-frame damping term from `length(wings) > 0` — a census of the model — and then indexed `wing_vel[:, point.wing_idx]` and `R_b_to_w[:, :, point.wing_idx]` with the point's own index. With a wing in the model and none on the point, that is column 0.

The same census ran a second time a few lines up, where `va_point_b` of every point that is *not* a wing node was expressed in wing 1's frame whatever wing the point named, with a comment saying "Use first wing for non-wing points". Between them they produced the incoherence Bart pointed at: a point's body-frame damping switched on or off according to whether some unrelated wing elsewhere in the model carried aero, since `sys.wings` holds only the aero-carrying bodies. Give a wing `AeroNone` and its neighbours' damping quietly disappeared.

### What I changed

One rule: `point.wing_idx` is the wing the point belongs to, and `0` means none. A point that names no wing gets `0`, whatever its dynamics type.

Every question about a point's wing is then answered by the point. `point_eqs!` no longer takes `wings` at all. `point_damping_accel` takes the wing velocities and decides for itself whether there is a wing to damp against, so the condition is written once rather than once in the caller and once again as a `vel_diff_w = nothing` sentinel. `va_point_b` is the point's own wing's frame, or zero. `va_frame_body` in the kernel backend is the point's `wing_idx` and nothing else, and `kinematic_wing_of` loses a bounds check the invariant now covers and a `!== nothing` test on a `KVec3` field that was never `nothing`.

Two checks hold the rule up, both at load: a wing reference has to name a wing that exists, and a station member has to name a wing at all — a station member *is* a wing's structural node, so "belongs to no wing" is not a state it can be in. The second sits after the constructor's "no wings provided, converting to STATIC" fallback rather than inside `mark_wing_nodes!`, or that fallback would become unreachable: it exists precisely to rescue wing nodes in a system with no wings.

### What this breaks

Three things change for a point that never named a wing:

- it loses its body-frame damping term, which was damping it against an arbitrary wing it had nothing to do with;
- its `va_b` is `0` rather than wing 1's frame — it feeds no aerodynamics, since only wing nodes carry per-point aero;
- a station member that names no wing, and a `wing_idx` naming a wing that does not exist, are now rejected by name at load instead of failing later or silently working against the wrong wing.

Nothing in the repo relied on the old default: every station member in every YAML and test names its wing, and the points whose value changes — `kcu`, `ground`, `nose` and their like — all carry zero body-frame damping, so their equations are the same equations. A model outside the repo that leaned on the default and set a non-zero `body_frame_damping` on a point it never gave a `wing` to would lose that term, and the changelog entry says so.

### A find I did not act on

`point_damping_accel` in the monolith and `body_frame_damp_accel` in `components.jl` are two spellings of the same damping physics, one per backend, held level by `test_backend_parity`. That is older and wider than this change and I left it alone; `rigid_wing_node` is likewise still written twice, once as the helper in `kernel_backend/assembly.jl` and once inline in `point_eqs.jl`. Worth a `cleanup:` PR that gives the structure itself the predicate, which is where it belongs.

### Verification

- [x] Reproduced first: `BoundsError: attempt to access ...BasicSymbolic... at index [:, 0]` from `point_eqs.jl` → `create_sys!` → `init!`
- [x] `test/test_point.jl` "Point without a wing" run against `main`'s `src/` in a fresh session: 7 passed, 5 failed, 1 errored — the error being the `BoundsError` above, and `Point(:free, zeros(3), DYNAMIC).wing_ref` evaluating `1 == 0`. Green on this branch: 18/18, with the file's own `Point Tests` 64/64
- [x] Affected tests green on the merge: `test_backend_parity` 3/3, `test_plate_aero` 69/69, `test_transform` 325/325, `test_wing_dynamics` 44/44, `test_deserialization` 10/10, `test_yaml_variables` 18/18, `test_yaml_bodies` 16/16
- [x] Local full suite: PASS — 3543/3543 in 31m, Julia 1.12.7, one cell of the matrix · GitHub CI: all seven checks green on `3d993d3a`, first attempt — including the 1.12 ubuntu monolith cell that needed a retry last round for #316's flake
- [x] Docs build clean (only the pre-existing size-threshold warnings of #301) · up to date with `main` · REUSE lint: not run locally, the tool is not installed on the box; CI's `reuse-lint` job covers it
- Benchmark: n/a — no hot-path change; one symbolic term fewer per wingless free point
- Risk: the load-time rejection of a station member with no wing. It cannot break a model that worked — such a model died on `bodies[0]` instead — but it is the one line that turns a previously-loading structure into an error.

### Scope

+122 / −60 across 9 files; `src/` alone is +42 / −58, which is the point of the change. No stack. Fixes #238.

Closes #238 · task `SymbolicAWEModels.jl-238`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@
   file untouched. Without it the script waits on a menu prompt.
 
 ### Fixed
+- A point no longer has to belong to a wing. `wing_idx: 0` is the documented "no
+  wing" sentinel, but the monolith backend decided whether a free point had a wing
+  frame to damp against by asking whether the *system* had any wing, so such a
+  point in a model that has a wing crashed model generation with a `BoundsError`
+  on index 0. It now asks the point, and a point whose wing is not among the
+  aero-carrying bodies keeps only its world-frame damping. A station member with
+  no wing — a wing's own structural node claiming to belong to no wing — is now
+  rejected with a message naming the point.
+
 - `AeroPressure` no longer writes one copy of a panel's live load per contour
   node. The symbolic component summed a separate `surface_node_forces` expression
   for each of the 210 nodes, so the panel's force residual and its pitching couple
--- a/docs/src/tutorial_yaml.md
+++ b/docs/src/tutorial_yaml.md
@@ -217,7 +217,7 @@ points:
 | `name` | String/Int | required | Point identifier (`idx` also accepted) |
 | `pos_cad` | [x,y,z] | required | Position in CAD frame [m] |
 | `type` | String | required | `STATIC`, `DYNAMIC`, or `BODY_STATIC` |
-| `wing_idx` | Int/nothing | 1 | Wing this point belongs to |
+| `wing_idx` | Int/nothing | 1 | Wing this point belongs to; `0` for none |
 | `transform_idx` | Int/nothing | nothing | Transform for initial positioning |
 | `body` | Ref/nothing | nothing | `BODY_STATIC`: body the point rides |
 | `joint` | Ref/nothing | nothing | `BODY_STATIC`: beam element the point rides |
--- a/src/generate_system/point_eqs.jl
+++ b/src/generate_system/point_eqs.jl
@@ -6,24 +6,17 @@
 """
     point_damping_accel(point, params, R_b_to_w, wing_idx, vel_w, vel_diff_w)
 
-Per-mass damping acceleration for a DYNAMIC point. Each frame's term is built
-only when its coefficient is set; a `nothing` coefficient keeps that term out of
-the equation entirely (no zero-valued parameter to prune). The body-frame term
-also needs a wing frame — pass `vel_diff_w = nothing` (point velocity relative to
-its wing) to skip it when no wing is available.
+Per-mass damping acceleration for a DYNAMIC point: a world-frame term against
+`vel_w`, plus a body-frame term in wing `wing_idx`'s frame against `vel_diff_w`,
+the point velocity relative to that wing. Pass `vel_diff_w = nothing` for a point
+that belongs to no wing, leaving only the world-frame term.
 """
 function point_damping_accel(point, params, R_b_to_w, wing_idx, vel_w, vel_diff_w)
-    accel = zeros(Num, 3)
-    if !isnothing(point.body_frame_damping) && !isnothing(vel_diff_w)
-        R = R_b_to_w[:, :, wing_idx]
-        coeff = params.points[point.idx].body_frame_damping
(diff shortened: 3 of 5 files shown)


1-Bort-1:
**Local full suite: PASS** (49 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - ubuntu-latest - x64 - monolith - pull_request` failed** on `6ba4af688`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34456101682/job/102802729878

CHANGELOG.md
```diff
@@ -3,6 +3,19 @@
 ## Unreleased
 
 ### Fixed
+- BREAKING: a point's `wing_idx` now means one thing — the wing it belongs to,
```
1-Bart-1: this is not a breaking change, it is a fix of a bug

1-Bart-1 (approved):

Issue #307, opened 2026-09-09
cleanup: CI's test matrix carries fail-fast: false


1-Bort-1:
`.github/workflows/CI.yml:38` sets `fail-fast: false` on the `test-with-code-coverage` matrix, so a failing job leaves every other cell running to the end. The matrix is ubuntu × windows × monolith/kernel backend at `timeout-minutes: 180`, and the kernel-backend cell alone is budgeted at ~100 minutes, so one broken commit spends hours of runner time re-proving the same failure.

rules/AGENTS.md §4.10 asks the opposite: "make the runner stop at the first failing test and the matrix at the first failing job, locally and in CI. A workflow carrying `fail-fast: false` is a `cleanup:` PR."

Dropping the line is the whole change. Worth checking first whether it was added on purpose to keep the windows cell reporting when ubuntu fails — the two do fail for different reasons often enough that someone may have wanted both answers; if so, this issue is the place to write that down instead.

Found while doing #302; out of scope there.

1-Bort-1:
No code change planned: the branch stays the two deletions in .github/workflows/CI.yml (fail-fast: false, and the exclude that excluded nothing). This turn only diagnoses the red macOS aarch64 cell on ba546100b and says on the thread whose bug it is. If it turns out a CI.yml edit can reach it, the fix rides here; otherwise the card stands and the finding is routed.

PR #320, opened 2026-09-11
Write and read a conforming structure document

### LINKED ISSUE #318: Write and read a conforming structure document
Write a `SystemStructure` out as a document conforming to `structure_schema.yml`, and
read one back.

There is an uncommitted `scripts/export_topology.jl` on at least one working copy that
does an early version of this. It should not be the basis: it hand-rolls a JSON emitter,
it predates the `Group` → `TwistSurface` → `Station` rename so it labels stations
`wing_sections`, and it emits connectivity only — no material, no bodies, no joints.
Delete it rather than port it.

What the schema expects, as blocks: `points`, `segments`, `stations`, `pulleys`,
`tethers`, `winches`, `bodies`, `elastic_joints`, `timoshenko_joints`. Every reference by
name; every component carrying its own geometry and material. Transforms are deliberately
not in the document — placement is not structure.

Two details that are easy to get wrong:

- **A wing is a body carrying aero.** `sys.wings` is the subset of `sys.bodies` whose
  `aero` is not `AeroNone`, and the document has one `bodies` block with an `aero`
  column rather than two blocks over overlapping sets.
- **`connectivity_sha` has an exact preimage**, documented in the schema: the point
  count, a semicolon, then each segment's endpoints as one-based row numbers into the
  points block, comma separated and semicolon terminated.

YAML and JSON should be two encodings of one model, sharing the loader — not two code
paths that drift.

**Done when** a V3 kite round-trips: `SystemStructure` → document → `SystemStructure`,
validating against the schema, with a golden-file test so the mapping stays true.

---

**Depends on** the schema, which exists. Watch for nonlinear stiffness laws: if the system being exported uses a `TubeRigidityLaw`, this is blocked on 1-Bart-1/awesIO#1, because the schema has only a placeholder for one today. Tracked by OpenSourceAWE/symawe.com#1.

1-Bort-1:
### TL;DR

A `SystemStructure` now writes out as a document conforming to awesIO's `structure_schema.yml` and reads back from one, with YAML and JSON as two encodings of the same document sharing one loader. A 2-plate kite round-trips `SystemStructure → document → SystemStructure → document` to an identical document, validated against a vendored copy of the schema and pinned by a golden file.

### What it does

`structure_document(sys)` builds the model: the `metadata` block plus one `headers`/`data` table each for `points`, `segments`, `stations`, `pulleys`, `tethers`, `winches`, `bodies`, `elastic_joints` and `timoshenko_joints`, every reference by name. `save_structure_document` and `load_structure_document` are the file ends — `.yml`/`.yaml` encode as YAML, `.json` as JSON, and both decode into the one `sys_struct_from_document` reader, so there is no second code path to drift.

A wing is a body carrying aero, so there is one `bodies` block with an `aero` column: `AeroNone` writes as null, anything else writes its type name, and the reader splits the rows on that column into the constructor's `wings` and `bodies`.

`connectivity_sha` hashes exactly the preimage the schema documents — the point count, a semicolon, then each segment's endpoints as one-based row numbers into the points block, comma separated and semicolon terminated. The test asserts it against the worked example in the schema's own text (`8;1,2;2,3;3,4;4,5;4,7;` → `d98529e2…`), so the preimage is pinned to the spec rather than to our implementation of it. The reader recomputes it from the document's own tables and refuses a document whose metadata does not describe them.

### The one design decision worth arguing about

**The document is the truth for everything it carries.** The schema's `bodies` block holds `mass`, `apparent_mass`, `inertia_principal`, `com_offset_b` and `pos_cad`, but its `points` block has no mass column — so a `SystemStructure` rebuilt from a document cannot re-derive a wing's mass from its points, the way `setup_wing_frame!` does when loading the structural YAML. The reader therefore constructs the system and then writes those five quantities back onto each body from its row, and re-runs `reinit!` so the world placement agrees with the restored CAD origin. Letting the constructor's derivation win instead would have made a wing's row a lie and broken the round trip.

The other inference: the schema has no wing column on a station, so a wing's stations are read as those whose points name that wing in the points block's `wing` column. That is the same membership the points already carry, so nothing is invented; a point's `wing` is written only for a wing node (`point.is_wing_node`), which keeps the column meaning what the schema's example means by it.

I deliberately did **not** append columns of our own, although the schema allows a writer to. A document with an `extra_mass` column would round-trip better and conform less; the golden file should document the schema, not our extension of it.

### What a round trip therefore loses

Worth knowing before this is built on, and each is a gap in the schema rather than in the mapping — written up with what I would do about them in 1-Bart-1/awesIO#3: a point's `extra_mass`, `area`, `drag_coeff` and the two damping vectors; the transforms that place the system in the world (deliberately out of scope per the schema); a wing's `R_b_to_c` and a particle wing's `origin`/`z_ref_points`/`y_ref_points`; a winch's Coulomb friction, viscous coefficient and total inertia (taken from `Settings` on read) and its motor model's own parameters; a pulley's debug `damping`/`brake`/`friction_epsilon`; a tether's `init_stretched_len` and `init_stretch_frac`; a beam-anchored point's joint reference.

Two of those are hard errors rather than silent losses, because a silent one would be worse: a `PARTICLE_DYNAMICS` (`KINEMATIC`) wing row is refused, since its body frame is fitted to reference points the document has no column for; and a winch or aero model whose name does not resolve to a zero-argument constructor is refused by name.

**On the tube-law risk flagged in the issue.** A callable `unit_stiffness` or joint rigidity raises, naming 1-Bart-1/awesIO#1. Nothing in `data/` uses one — `TubeRigidityLaw` is only reachable through the exported `tube_bending_law`/`tube_torsion_law`, which no shipped system calls — so this turn is not blocked, and the failure is loud the day a system does.

### Two things about the setting

There is no V3 kite in this repository; it lives in `V3Kite.jl`. The in-repo kite is `data/2plate_kite`, whose VSM wing is literally named `V3_Kite`, and that is the system the golden document describes. A second, programmatic case covers what the 2-plate kite has none of: plain bodies, an elastic joint, a Timoshenko joint and a body-anchored point.

`scripts/export_topology.jl` is not in this worktree — it exists only on someone's working copy — so there was nothing to delete here. Nothing in this change derives from it.

### Dependencies, and why the manifests moved

`JSON` is a new root dependency: the issue asks for JSON as a first-class encoding, and hand-rolling an emitter is the thing the old script got wrong. Both tracked manifests already carried it through the workspace, so against `main` the minimal resolve adds exactly one package — `JSONSchema`, a new **test** dependency for validating against the vendored schema — and moves `project_hash`, `JSON` into the root project's deps line, and the `SymbolicAWEModels` entry's own `version`, which every release since v0.16.0 has left stale in both files and which now follows `Project.toml` to `0.18.0`. Each `.default` was re-seeded from `main`'s and re-resolved under its own Julia, 1.12 and 1.13; no package version moved. `YAML` is the other new test dependency, and was already a root dep.

`test/data/structure_schema.yml` is a byte copy of `src/awesio/schemas/structure_schema.yml` from `1-Bart-1/awesIO@structure-schema`, MIT, annotated as such in `REUSE.toml`. awesIO is a Python package, so there is no way to depend on it from Julia; a vendored copy is how the contract gets compiled against.

### Where I would push back

The golden document is 614 lines for 15 points, because YAML.jl's writer only emits block style — one scalar per line, where the schema's hand-written example uses flow rows. Nothing in the mapping causes it and I did not want to hand-roll an emitter to fix it; a real kite's document will be a few thousand lines and should probably be read as JSON.

`docs/src/private_functions.md` grows by 30 entries because this repo's `makedocs` checks every docstring, not just exported ones. That page was already over Documenter's warn threshold — #301 — and this makes it slightly worse without changing which pages warn.

### Verification

- [x] Reproduced first: n/a — new feature, nothing to reproduce
- [x] `test/test_structure_document.jl`: **19/19 on this head**, all seven testsets (juliaserver, exit 0). The golden test is not vacuous: changing one `pos_cad` component in `test/data/2plate_kite_structure.yml` from `2.3` to `2.35` turns it red (1 pass / 1 fail), and it is green again on the restored file
- [x] `test/test_helpers.jl` 9/9 — the manifest-staleness check, which is what the re-resolve could have broken
- [x] `test/test_tether_init.jl` 58/58 on this head — what `main`'s #349 changed, the `reinit!` the reader calls
- [x] Docs build clean and REUSE lint clean: CI's `Documentation` and `reuse-lint` jobs green on bda8d574 · up to date with `main` (5949e099, #349)
- [x] Local full suite (`agent ci-local`, one cell of the matrix): **PASS** in 33 min, exit 0 (Julia 1.12.7, 3593/3593), on bda8d574 · GitHub CI: run 35118822675 on bda8d574 **PASS on all seven jobs**, Windows and macOS included
- Benchmark: n/a — a file writer, not a hot path
- Risk: the reader overwrites each body's mass, inertia, COM offset and CAD origin after construction and re-runs `reinit!`. If some later step is added that derives those *after* `reinit!`, the round trip would silently start losing them; the golden test is what would catch it.

### Scope

+1819 / -6 across 14 files, of which 1097 lines are data — the vendored schema (483) and the golden document (614) — leaving 478 lines of `src/structure_document.jl`, 138 of test and 53 of docs. The source is long for one file because the format has nine blocks, each a declarative column list in both directions; splitting the writer from the reader would leave neither independently testable, since the round trip is the test. `CI.yml` still carries `fail-fast: false` (#307), untouched here.

Closes #318 · task `SymbolicAWEModels.jl-318`


Manifest-v1.11.toml.default
```diff
@@ -908,6 +908,16 @@ git-tree-sha1 = "7204148362dafe5fe6a273f855b8ccbe4df8173e"
 uuid = "692b3bcd-3c85-4b1f-b108-f13ce0eb3210"
 version = "1.8.0"
 
+[[deps.JSON]]
```
1-Bort-1, line 911: MAJOR: This default manifest was updated for JSON/StructUtils but omits the new test-only dependency JSONSchema, while the v1.12 default includes it; the pinned 1.11 environment cannot run the new tests without re-resolving.

1-Bort-1 (commented):

**Verdict: REQUEST CHANGES** · 2 inline, 0 off the diff

**Good**

- Single feature scope: SystemStructure ↔ awesIO structure document round-trip matches the issue and keeps one idea per PR.
- YAML and JSON share one loader: save_structure_document dispatches by extension and both call sys_struct_from_document.
- Schema validation and the worked-example connectivity_sha test pin the contract, not just our implementation.
- New public API functions have concise docstrings and are added to docs/src/exported_functions.md.
- Hard errors for KINEMATIC wings and unresolvable model names, with issue references.
- CHANGELOG, REUSE.toml annotations, and focused tests are present; no unrelated source churn.

**Not good**

- `Manifest-v1.11.toml.default:911` — This default manifest was updated for JSON/StructUtils but omits the new test-only dependency JSONSchema, while the v1.12 default includes it; the pinned 1.11 environment cannot run the new tests without re-resolving.
- `test/data/2plate_kite_structure.yml:1` — Committing the writer's output as a golden fixture violates the generated-files rule; round-trip and schema tests already prove correctness, and any pinned expected output should be a hash or generated at test time.
- src/structure_document.jl uses decorative # === section banners; the rubric wants a block that needs a header to become a function.
- read_point/read_segment/read_station/read_pulley/read_tether/read_winch/read_body/read_elastic_joint/read_timoshenko_joint lack docstrings.
- Several lines in src/structure_document.jl exceed the 92-character wrap limit.
- docs/src/private_functions.md grows by 30 entries and worsens the pre-existing #301 size-threshold warning.
- test_structure_document.jl header comment says regenerate by saving to data/2plate_kite_structure.yml, but the file lives in test/data.

test/data/2plate_kite_structure.yml
```diff
@@ -0,0 +1,614 @@
+metadata:
```
1-Bort-1, line 1: MAJOR: Committing the writer's output as a golden fixture violates the generated-files rule; round-trip and schema tests already prove correctness, and any pinned expected output should be a hash or generated at test time.

1-Bort-1:
**Local full suite: PASS** (21 min, Julia 1.13.0, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `912986c72`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34961273854/job/104355182926

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `ab72843b7`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34970450635/job/104385700328

1-Bart-1:
Fix the failing windows test

1-Bort-1:
Fixed in 78f9082f, which merges #348's fix; the Windows cell is green on it in run 35079969394, all seven jobs pass.

1-Bort-1:
**CI: `Julia 1.13 - ubuntu-latest - x64 - monolith - pull_request` failed** on `bb21f0758`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/35088429341/job/104768705617

src/structure_document.jl
```diff
@@ -0,0 +1,478 @@
...
+Render `sys` as a document conforming to `structure_schema.yml`: one
+`headers`/`data` table per block, every reference by name, every component
+carrying its own geometry and material. Returns nested `OrderedDict`s and
+`Vector`s, which [`save_structure_document`](@ref) encodes as YAML or JSON.
+
+The schema describes structure only, so the transforms that place the system in
+the world, the live state, and the point masses and drag properties it has no
+column for are left out.
+"""
+function structure_document(sys::SystemStructure; name::AbstractString=sys.name,
```
1-Bart-1, line 111: I would assume this kind of stuff would need to be in the KiteGeometry.jl package

src/structure_document.jl
```diff
@@ -0,0 +1,478 @@
...
+    rows = Dict(block => document_rows(doc, block) for block in keys(DOCUMENT_HEADERS))
+    check_connectivity(metadata, rows["points"], rows["segments"])
+    resolved_set = isnothing(set) ? load_settings("base") : set
+
+    wing_rows = filter(is_wing_row, rows["bodies"])
+    stations = Station[read_station(row) for row in rows["stations"]]
+
+    sys_struct = SystemStructure(string(metadata["name"]), resolved_set;
+        points = Point[read_point(row) for row in rows["points"]],
+        stations,
```
1-Bart-1: why is stations not read

src/structure_document.jl
1-Bort-1: Yes, and it is already scheduled: 1-Bart-1/Agents#294's plan has a unit whose whole subject is moving `src/structure_document.jl` into OpenSourceAWE/KiteGeometry.jl, and it waits on OpenSourceAWE/KiteGeometry.jl#22 to put `SystemDefinition` and the YAML loader there first. KiteGeometry.jl is empty today and nothing depends on it, so the encoder stays beside the type it encodes until then; no second issue filed.

src/structure_document.jl
1-Bort-1: It is read — cf52c397 reads it inline in the constructor call, where it was hoisted a line above and passed by the `stations,` shorthand, which is what made it the one block that did not look read.

PR #291, opened 2026-09-05
next_step! should gain a kwarg that makes it warn on failing vsm solve

1-Bort-1:
### TL;DR

`next_step!(sam; vsm_warn_on_fail=true)` turns a failed VSM solve into a warning and flies the wing on with the circulation, the angles of attack and the frozen forces of its last converged solve, instead of ending the run. The failure itself is `VortexStepMethod.SolveFailure`, raised by VSM's own `solve!` under `throw_on_fail`, which is what the bump to VortexStepMethod v5.1.0 in this diff is for.

### What a failed solve used to cost

`safe_vsm_solve!` zeroed the circulation on a failure and every mode's refresh threw an `AssertionError`, so one bad step ended the simulation and the next solve started from zero rather than from the last good circulation. Both halves are fixed. The last converged circulation and both angle-of-attack distributions are put back before the failure propagates — `solve!` has already overwritten them with the diverged ones by the time anyone can look — and `vsm_warn_on_fail` downgrades the failure to a warning. `vsm_interval` is untouched, so the next scheduled update solves again; the assertions on an already corrupted frozen state stay fatal, because a wing whose point forces are NaN cannot be flown on.

`AeroLinearized` is the one place where "nothing frozen is written before the solve can fail" is not true, and `docs/src/vsm_coupling.md` says so: a solve that fails during the Jacobian pass keeps the baseline coefficients of the operating point it just refreshed, next to the Jacobian columns the pass had not reached.

### Where the error lives, and the pin that reaches it

Review said the error should exist and be thrown in VortexStepMethod, not here. It does: OpenSourceAWE/VortexStepMethod.jl#282 added `SolveFailure` and `solve!(...; throw_on_fail=true)` and took this package's Dual-aware `finite_full` with it — whether a solve came back usable is VSM's question, not a caller's. This branch deletes `VSMSolveFailure` and `finite_full`, the three `if !safe_vsm_solve!(...) throw(...)` blocks collapse to plain calls, `safe_vsm_solve!` catches, restores and rethrows, and `warn_or_rethrow` names the mode and the wing that VSM's own message cannot. `src/aero_modes/common.jl` gets shorter by doing it.

[...]

PR #286, opened 2026-08-25
Fix roll, pitch and yaw: NED formula was applied to an ENU quaternion

1-Bart-1:
`wing.Q_b_to_w` is ENU, but `update_sys_state!` computed `ss.roll`/`pitch`/`yaw` from it with a NED aerospace formula. The result was not the angles any sensor or controller reports.

Measured, kite at zenith with the nose north and wings level:

| | roll | pitch | yaw |
|---|---|---|---|
| before | 0° | 0° | **−90°** |
| after | 0° | 0° | 0° |

`KiteUtils.euler_ks` converts the `KA` attitude to `KS` first, which is the convention roll, pitch and yaw are defined in — against NED, because that is what the sensors deliver.

`ss.orient` needs no conversion: KiteUtils 0.13 stores `KA`, which is what `Q_b_to_w` already is. SAM was the package that already had this right.

**Depends on OpenSourceAWE/KiteUtils.jl#130** for `euler_ks`; compat bound raised to `0.13`, so this cannot merge before that release.

Three reference attitudes are pinned in KiteUtils' `test-frames.jl`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #322, opened 2026-09-12
Julia 1.13: SymbolicAWEModels.jl

### LINKED ISSUE #321: Julia 1.13: SymbolicAWEModels.jl
The shared procedure. The heaviest of these: `CI.yml` carries a matrix plus four
`include:` entries naming 1.11 and 1.12 individually, `setup-test.yml` is on
`'1'`, both `Manifest-v1.11` and `Manifest-v1.12` defaults are tracked, and
`bin/install`, `bin/run_julia`, `bin/setup_env`, `bin/install_jetls` and
`bin/create_model_archive` all name a Julia.

Symbolic compilation makes the resolve and the test run long. Budget for it and
do not let a slow first run look like a hang.

Depends-On: https://github.com/OpenSourceAWE/KiteUtils.jl/pull/139

Unit `symbolicawe` of plan `plans/julia-1-13.md`, tracked in 1-Bart-1/Agents#282.
https://github.com/1-Bart-1/Agents/blob/e955329df93f2e20b94eeef6ff818c3a879989a1/plans/julia-1-13.md

1-Bort-1:
### TL;DR

CI's one 1.11 cell becomes a 1.13 cell, `Manifest-v1.11.toml.default` is replaced by one resolved under Julia 1.13, and every `bin/` script that offered 1.11 now offers 1.12 and 1.13. This is the `symbolicawe` unit of `plans/julia-1-13.md`, so the repo tests and installs the pair the fleet is moving to.

### Which of the five steps this repo had

**The matrix.** `CI.yml` names a minor in five places — the base matrix `version: ['1.12']` and four `include:` entries — and exactly one of them, the ubuntu entry carrying `prefix: xvfb-run -a`, was on 1.11. That entry is the whole workflow diff. 1.13 moved into it rather than into the base matrix because a bare ubuntu cell there would run the GLMakie tests with no display; this way 1.13 inherits the xvfb prefix 1.11 had. The jobs are now 1.12 on ubuntu (coverage), windows and macOS-aarch64, 1.12 again for the kernel backend, and 1.13 on ubuntu. `setup-test.yml` is on `'1'`, which names no minor, so it is untouched.

**The compat line.** `julia = "1.11, 1.12"` → `"1.11, 1.12, 1.13"`. An append, so the admitted range is unchanged and 1.11 is not dropped — it is only no longer a version the repo's own tooling installs or tracks a manifest for. On a 1.11 machine `bin/install` now exits 1 saying so, which is the honest answer.

**The manifests.** `Manifest-v1.11.toml.default` deleted, `Manifest-v1.13.toml.default` added by the minimal resolve of `AGENTS.md` §2 (the 1.12 default seeded under 1.13, `Pkg.resolve()`, copied back), `Manifest-v1.12.toml.default` re-resolved under 1.12. Between the two tracked defaults only stdlib and JLL versions differ — `SHA`, `LinearAlgebra`, `SparseArrays`, `Pkg`, `LibCURL` and nine JLLs — plus `manifest_format` 2.0 → 2.1. No package version moves, which is the point of seeding rather than updating.

**The scripts.** `bin/install` (usage, the juliaup install hint, the menu and its fallback, the accept-list, the stale-manifest sweep), `bin/update_default_manifests`, `bin/install_jetls`, and the usage comments in `bin/create_model_archive` and `bin/setup_env`. Where a script picks one version rather than a pair it now picks 1.13, the newer of the two. Two named a Julia and did not move: `bin/run_julia` reads the minor off `julia --version` and its only literal is a `>= 1.12` guard for kaimon, which is a bound rather than an enumeration; `bin/create_sys_image` takes the channel as an argument.

**The test.** `test_helpers.jl` asserts the mtime of each tracked `.default` by name, so the pair it names follows.

### Dropping 1.11 leaves dead code in bin/install

`Project.toml` has a `[workspace]`, which is a 1.12 feature, and `bin/install` had two version-conditional blocks around it: on 1.11 it seeded `examples/` and `test/` from the root manifest, and on 1.12 it deleted the sub-project manifests again after each precompile. Both supported versions are workspace versions now, so the seeding branch goes and the sweep becomes unconditional — and moves out of the loop that was running it once per sub-project. The stale-0-byte sweep in `_prepare_manifests` listed one version per line; it is a glob now, so the next Julia costs it no edit. `.gitignore` had the same accretion — live manifests listed for 1.10, 1.11 and 1.12, plus two `.bak` paths the `*.bak` line already covers — and is one `Manifest-v1.*.toml` glob.

### Two merges with main, and no KiteUtils dependency left

#291 landed while this was open and moved `VortexStepMethod` to 5.1.0, which is a pin in both tracked defaults — a conflict against a branch that deletes one default and adds another. Neither side was taken: the minimal resolve was re-run on top of main's pins, the 1.12 default seeded from main's and resolved under 1.12, then the 1.13 default seeded from that result and resolved under 1.13. Both carry VortexStepMethod 5.1.0, and the 1.13 default differs from the pre-merge one in exactly three lines — that version, its tree hash, and the `project_hash` the compat edits move.

The v0.17.0 release (#325) then took the changelog's `## Unreleased` header for itself and moved that section's `### Changed` block below its `### Fixed` — which is where the 1.13 bullet was, so the two rewrote the same lines. main's structure is taken whole and the bullet goes into a fresh `## Unreleased` section above the release: it is not part of v0.17.0, and `bin/release` reads the top `## ` header's first field as the version going out, so `Unreleased` on top is what between-releases looks like. `Project.toml` merged clean, main's version bump beside this branch's two compat lines, and that bump leaves `project_hash` alone — `Pkg.Operations.is_manifest_current` is true under both 1.12 and 1.13 against the merged project, so neither default needed re-resolving and no manifest line moved for the release.

OpenSourceAWE/KiteUtils.jl#139, the branch this unit was speculating against, has merged, so the override is gone and this is no longer a draft. It needs no KiteUtils release: the General registry compresses the released 0.12.2's `julia = "1.10, 1.11, 1.12"` to the range `1.10.0 - 1`, so the registered version resolves under 1.13 unchanged — which is what the 1.13 default now pins, with no `[sources]` entry and no git rev anywhere in the diff.

### Found on the way

`Pkg.resolve()` under 1.13 warns `Ignoring incompatible compat entry SHA = "0.7.0" ... SHA is a non-upgradable standard library with version 1.0.0 in the current Julia version`. The SHA stdlib went 0.7.0 → 1.0.0 in 1.13, so the bound this package states is silently dropped there. `SHA = "0.7.0, 1"` — what Pkg's own message asks for — rides here as a one-line 1.13 fix in a file the change already opens. It is also why `Manifest-v1.12.toml.default` is in the diff at all: compat feeds `project_hash`, and that one line is its entire change.

The v0.17.0 merge showed #305 from the other side. Merging rewrote `Project.toml`, which put its mtime past both `.default`s, and `test_helpers.jl` compares exactly those mtimes — so a merge that re-resolves nothing turns the check red on manifests that are not stale. The byte-identical live manifests were copied back over the `.default`s (step 4 of the minimal resolve) to refresh the mtimes; `git status` stays clean on both files. Nothing here fixes the proxy — #305 is the thread for that.

Two things noted and left alone. `CI.yml`'s `exclude:` names `os: macOS-latest`, which is not in the base matrix's `os` list, so it excludes nothing and has not since macOS moved into an `include:`. And `Artifacts.toml.default` still points at `models_v1.10` and `models_v1.11` tarballs from the v0.5.0 release, which nothing in the repo reads — a 1.13 entry would be meaningless before that is decided. `fail-fast: false` stays for #307, as the plan asks.

### Verification

- [x] Reproduced first: with `Manifest-v1.11.toml.default` gone, `test_helpers.jl`'s own assertion on it evaluates `0.0 >= 1.7892151388947225e9` and fails; the 1.12/1.13 pair it names now passes 9/9
- [x] Local full suite under Julia 1.13, monolith, on `33200d71` and against the registered KiteUtils 0.12.2: PASS, 3519/3519 in 27m10 — which includes #291's own `test_aero_modes` 809/809, on 1.13 and VortexStepMethod 5.1.0 for the first time. `agent ci-local` ran the 1.12 cell there: PASS, 45 min
- [x] GitHub CI on `33200d71`: all seven checks green — the new ubuntu/1.13 cell in 41m, ubuntu/1.12 monolith, ubuntu/1.12 kernel, macOS-aarch64/1.12, windows/1.12, Documentation and reuse-lint
- [x] The v0.17.0 merge (`45e72e5e`) changed only changelog prose and the version string, so it was re-verified rather than re-suited: `test_helpers.jl` 9/9 under 1.12 and 9/9 under 1.13, and `test_pulley.jl` 82/82 under 1.13 on the model cache the bump re-keys (`model_v0.17.0_jl1.13_pulley_test_…`)
- [x] GitHub CI on `45e72e5e`: six of seven green — the new ubuntu/1.13 cell in 46m7, ubuntu/1.12 monolith in 1h12, ubuntu/1.12 kernel in 1h19, macOS-aarch64/1.12 in 1h04, Documentation and reuse-lint; windows/1.12 was still running when this was written, as it was the last to land on `33200d71` too (1h25, pass). `agent ci-local` was not repeated on this head — the box's memory watchdog reaped it and the Julia session with it at 40/62 GB used, and starting a second full matrix cell there would be taking the box off the agents also on it
- [x] Up to date with `main` · docs unchanged (no public symbol moves) · REUSE lint green on CI (`reuse` is installed on neither this box nor as docker)
- Benchmark: n/a
- Risk: the 1.13 default was produced by resolving the 1.12 one under 1.13, so it keeps the 1.12 pins wherever 1.13 allows them. Deliberate — it makes the diff between the two files the Julia-version delta and nothing else — but the 1.13 tree is not independently the newest resolvable one.

### Scope

+1800 / −115 across 12 files, of which 45 / −44 is hand-written; the rest is `Manifest-v1.11.toml.default` rewritten as the 1.13 one plus a single `project_hash` line in the 1.12 one. Net −3 lines of script and workflow. No stack: #320 and #286 also touch `Project.toml` and the tracked defaults, but neither touches the `julia`, `SHA` or matrix lines this changes, so whichever lands second only re-resolves.

Closes #321 · task `SymbolicAWEModels.jl-321`

Closes #321 · task `SymbolicAWEModels.jl-321`


1-Bort-1:
**Local full suite: PASS** (74 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

PR #332, opened 2026-09-14
Delete the wing's tether wrench: two constructor-zero fields, their call sites, and the panel that plotted them

### LINKED ISSUE #323: SysState drops tether_induced_force and tether_induced_moment: three call sites go with them
KiteUtils 0.13 removes `SysState.tether_induced_force` and `SysState.tether_induced_moment` (OpenSourceAWE/KiteUtils.jl#138, in OpenSourceAWE/KiteUtils.jl#140). Nothing ever filled them here, which is why they went: `Body.tether_force` and `Body.tether_moment` are declared at `src/system_structure/rigid_body.jl:116-117`, but the wing readout map at `src/model_management.jl:106-122` syncs `aero_force_b` and `aero_moment_b` and not those two, so the value copied into the state has not moved since construction.

Three call sites break when that release lands:

- `src/symbolic_awe_model.jl:387-388` — `ss.tether_induced_force .= wing.tether_force` and the moment line beside it. Delete both.
- `src/system_structure/utilities.jl:1042-1043` — the restore path copies the two columns back into the wing. Delete both; they restore a constant.
- `ext/SymbolicAWEModelsMakieExt.jl:1719-1727` — the `plot_tether_moment=true` panel plots `sl.tether_induced_moment[i][2]`, which is that same constant. Removing the keyword is a user-visible change of this package, so it is yours to decide; the alternative is to plot the quantity properly, summing the `spring_force` entries of the segments attached to the wing.

`Body.tether_force` and `Body.tether_moment` have no writer left afterwards, so they go with the call sites unless something is meant to fill them.

The two packages are not coupled through this: KiteUtils can release first, and this repo's CI installs the registered version until it does.

1-Bort-1:
### TL;DR

`Body.tether_force` and `Body.tether_moment` were written once each, as `zeros(KVec3)` by their own constructor, and no readout in either backend ever moved them again — so the `SysState` columns they fed, `tether_induced_force` and `tether_induced_moment`, carried a constant for the life of every run. KiteUtils 0.13 removes those two columns, and the two fields, their two call-site pairs and the Makie panel that plotted the moment go with them.

### What was actually there

The two fields had two writers between them, and both are a constructor: `Body(name; …)` and `Wing(name, stations, …)`, where each is the fifth and sixth of a six-`zeros(KVec3)` run. The wing readout map syncs `aero_force_b` and `aero_moment_b` and stops there; the kernel backend's `WingAeroReadout` carries the same two and no more; `sys.state_vars` does not list them. Nothing else in the package assigns to either.

So three call sites were a closed loop carrying zero: `update_sys_state!` copied the zeros into the two `SysState` columns, `update_from_sysstate!` copied them back out of a loaded log into the same fields, and the Makie extension's `plot_tether_moment` panel drew `sl.tether_induced_moment[i][2]` — a flat line on the axis. The KiteUtils removal is what made the loop visible; it was already dead.

### Why the panel is removed rather than repaired

The panel method does receive the `SystemStructure`s, so it could find the segments hanging off a wing and it is not blocked for want of topology. What blocks it is the quantity: `SysState` logs `spring_force` as one scalar magnitude per segment, so a moment would have to be rebuilt inside the plotting extension out of logged point positions and the wing pose. That is a second implementation of physics the model itself does not compute, which `CLEAN_CODE.md` §2 calls the defect rather than the fix — and the column it would then be logged into is the one KiteUtils 0.13 takes away.

A wing's tether-induced wrench has no source in this package, so there is nothing to put in the panel's place. If it should have one, that is a feature with its own thread and its own order: an observable on the wing, an entry in both backends' readouts, a `SysState` column to carry it, and only then a panel.

### No compat bump rides here

The deletions are correct against the pinned `KiteUtils = "0.12.2"` as well — the two columns still exist there, they are simply no longer written — so this branch merges on its own and does not wait on a KiteUtils release or on OpenSourceAWE/KiteUtils.jl#140. Moving this repo onto KiteUtils 0.13 is a separate change: that release also renames `aero_force_b`/`aero_moment_b` to the `_KA` names, which is a different diff in the same two functions.

### Two finds, neither taken here

`ext/SymbolicAWEModelsMakieExt.jl:1825-1915` is a 90-line commented-out debugging block inside the `plot_gk` panel, three of whose lines read `sl.tether_induced_force`. It is dead code and should go (`CLEAN_CODE.md` §6), but #329 is open and running on that exact panel, so deleting it here would race that branch. It belongs in #329 or in a `cleanup:` PR.

Nothing in `test/` calls the multi-panel `plot(::SystemStructure, ::SysLog)` at all — `test_makie_extension.jl` exercises the 3D scene and the recorder, not the panels. That is why #329 exists, and it is why the panel check below is a probe rather than a test; adding that coverage belongs on #329, whose whole subject is a panel that throws.

### Verification

- [x] Reproduced first: n/a in the usual sense — a field nothing writes has no failing run. Standing in for it is the writer census above, run over `src/`, `ext/` and `test/`: two constructors, zero readouts, one reader.
- [x] `test/test_beam_replay.jl` — the `update_sys_state!` → `save_log` → `load_log` → `update_from_sysstate!` round trip, on six plain `Body`s: **7/7 pass**, 7m27s (juliaserver, Julia 1.12.7).
- [x] `test/test_wing.jl` — the `Wing(…)` constructor on both dynamics types, and `validate_sysstate_roundtrip` through `test_init!`: **171/171 pass**, 6m03s.
- [x] `test/test_makie_extension.jl`: **15/15 pass**, 6m26s. `test/test_deserialization.jl`: **10/10 pass**, 14.1s.
- [x] The edited panel method run directly on a 2-plate rigid wing with six panels on, `plot_aero_moment` among them, returns a `Figure`; `plot_tether_moment=true` now raises `MethodError: no method matching plot(…)`.
- [x] Docs build clean (`docs/make.jl`): no errors, and the three size-threshold warnings are the ones #301 is open about. Working tree clean afterwards. Up to date with `main` (558b27c7) — one commit on top, no merge needed.
- [ ] Local CI mirror (`agent ci-local`, one matrix cell): started at the push, not waited on. GitHub CI is the authority, and its test workflow triggers on the pull request rather than on the branch, so it starts with this card.
- REUSE lint: **not run** — `bin/reuse_lint` reports `Could not run REUSE lint. Install one of: pipx + reuse, or python package reuse.` `bin/jetls` likewise: `jetls: command not found`. No file gains or loses a licence header here.
- Benchmark: n/a — nothing on a hot path changed.
- Risk: the constructor arity. Both `Body{…}(…)` calls lose one positional argument each, and a miscount would shift every field after `aero_moment_b` silently rather than error. Counted at 57 against the struct's 57 and checked field by field; `test_beam_replay.jl` covers the first constructor and `test_wing.jl` the second, both of which read `elevation`, `heading` and `turn_rate` back out.

### Scope

+12 / −33 across 7 files. Of the 12 insertions, 7 are the changelog bullet and 3 the reflow of the opt-in panel list in `docs/src/exported_functions.md`; the remaining 2 are the two shortened `zeros(KVec3)` constructor lines. No new code.

Closes #323.

Closes #323 · task `SymbolicAWEModels.jl-323`


1-Bort-1 (commented):

**Good**

- Body field removal is complete: `rg` over src/ext/test finds no remaining reader or writer of `Body.tether_force`/`tether_moment` — the only surviving `tether_force`/`tether_moment` hits are the live station-level fields (`src/model_management.jl:86-87`, `src/kernel_backend/state.jl:465-466`), which this diff correctly leaves alone.
- Constructor arity checked by hand against the struct: the aero/wing block is `wind_disturb, drag_frac, va_b, wind_vec, aero_force_b, aero_moment_b`, so both shortened lines (rigid_body.jl:294, wing.jl:235) now pass exactly 4 `zeros(KVec3)` where 6 stood, and the 7 scalars / 4 trailing entries after them still line up.
- The dead-field claim holds at the source: the kernel `getter.aero` readout writes only force/moment/apparent/wind (`src/kernel_backend/state.jl:470-474`) and `model_management.jl` syncs only `aero_force_b`/`aero_moment_b` — nothing ever moved the two deleted fields off their constructor zeros.
- Feature removal takes its callers with it, per §2: the two `update_sys_state!` lines, the two `update_from_sysstate!` lines, the panel body, its keyword default and its docstring line, plus the docs panel list and a CHANGELOG entry — no orphan keyword or doc reference left (`rg plot_tether_moment` is clean).
- Deletion-over-repair is argued on the right grounds and the alternative is named in the card rather than half-built; no speculative replacement observable, column or panel was added, and the diff adds no code at all.

**Not good**

- `ext/SymbolicAWEModelsMakieExt.jl:1903` keeps a commented-out line reading `sl.tether_induced_force` inside `plot_gk`; the card defers it to #329 to avoid racing that branch, which is defensible, but it does leave a reference to a column this PR declares gone.
- The CHANGELOG bullet spends four of its seven lines on why the fields were dead — house style in this file is comparably discursive, so this is consistent rather than wrong, but the rationale properly belongs in the PR card (§5).
- The docs reflow at `docs/src/exported_functions.md:162-164` rewraps three lines to drop one name; harmless, but it makes the one real deletion harder to spot in the diff than a single-line edit would.

PR #333, opened 2026-09-14
Draw the steering-gain panel instead of throwing UndefVarError

### LINKED ISSUE #329: plot(..., plot_gk=true) throws UndefVarError: cs_over_us_vec

1-Bort-1:
### TL;DR

`plot(..., plot_gk=true)` threw `UndefVarError: cs_over_us_vec` on a name that has not existed since v0.12.0, so the panel could not be asked for at all. The line went with the rest of the debugging session it belonged to, and what survives is the steering gain and nothing else.

### What was there

The `if plot_gk` branch computed `g_k` in six lines and then spent a hundred and twenty on a diagnosis: twelve unconditional `@info` calls printing every intermediate on every call, a `window_mean` closure averaging them over the last fifty seconds, and ninety commented-out lines working through an "alpha mystery" — three ways of computing α, a lift/drag decomposition, and the same again for the tether. `cs_over_us_vec` was one of the averaged quantities; `calculate_cs`, which produced it, went in v0.12.0 with the rest of the V3-specific analysis code, and this reference was left behind. Nothing covered the branch, so nobody found out.

I deleted the diagnosis and kept the panel. `g_k = ψ̇ / (v_a u_s)` is the steering gain, a quantity worth a panel; what was wrapped around it was a session, not a feature. Deleting the whole branch was the other option the issue offered, but that removes two public keyword arguments to fix a crash in one of them.

### What the panel does now

`gk_ylims` reaches the axis. It was accepted at `:1333` and read nowhere — the panel pushed no `ylim` key and the renderer applies limits only under that name. The default stays `(0.0, 15.0)`; the docstring said `(0.0, 10.0)` and now agrees with the code.

The heading rate was computed twice in this file, the same ten lines in this panel and in `plot_yaw_rate`. Both now call `compute_heading_rate`, which is not a third copy of the unwrapping either: `wrap_to_pi` already exists in `src/system_structure/transforms.jl`, and a rate over wrapped steps is what the unwrap loop was computing the long way. Over 2000 random wrapped-heading series the two agree to 4e-12; they differ only on a step of exactly π, where the old loop reads +180 °/s and `wrap_to_pi` reads −180 °/s, a half turn whose sign is arbitrary in both.

### What I did not change, and why

**`turn_radius_ylims` is broken the same way, one screen up.** That panel does push its limits, under the key `ylims`, and the renderer reads `ylim` — one letter, `:1567` against `:2381`. Its documented default `(0.0, 100.0)` disagrees with the signature's `(0.0, 70.0)` too. Giving a panel limits it has never had is a behaviour change on a panel this PR has no other reason to touch, so it is yours to call: say the word and it is two characters and a docstring line.

**`SysState.steering` has no agreed unit.** This panel reads it as a percentage — `abs(u_s) > 1.0`, then `u_s / 100` — and the neighbouring `plot_us` panel labels it `[\%]`. KiteUtils documents the field as `[-1..1]`. This repo's own `update_sys_state!` writes neither: `src/symbolic_awe_model.jl:363` puts `rad2deg(twist_angles[end] - twist_angles[1])` there, degrees of twist difference. All three readings give different `g_k`, and on a log honouring KiteUtils' range every value is `NaN`. I left the formula exactly as it was; which one is right is a question about what callers write into the log, not about this crash.

**The two callers of `compute_heading_rate` stamp it on different samples.** `plot_yaw_rate` uses `sl.time[1:end-1]`, `plot_gk` uses `sl.time[2:end]`, so the same quantity is drawn one sample apart on the two panels; `plot_yaw_rate` also clamps jumps over 200 °/s and `plot_gk` does not. Both predate this branch and both are visible changes to what is drawn, so they stayed out of it.

**`aoa_ylims` was documented and does not exist.** It appeared in the docstring and on `docs/src/exported_functions.md` but is not a keyword of the method — `plot(sys, log; aoa_ylims=(0, 1))` is a `MethodError`. I removed both mentions rather than leave a documented argument that throws; `ylims=Dict(:aoa => (low, high))` is the mechanism that works, and the AoA panel already reads it. Four more keywords of this method (`plot_us`, `plot_sideslip`, `plot_kite_vel`, `plot_wind`) are in neither the docstring nor the docs page; documenting them is a docs pass, not this diff.

### Verification

- [x] Reproduced first: `UndefVarError: cs_over_us_vec not defined in SymbolicAWEModelsMakieExt`, at `ext/SymbolicAWEModelsMakieExt.jl:1836`
- [x] `test/test_makie_extension.jl` red before — `gk panel draws and honours gk_ylims | 1 errored`, the other 15 assertions passing — green after: 18/18, exit 0 (juliaserver, GLMakie headless)
- [x] `compute_heading_rate` equals the loop it replaces to 4e-12 over 2000 random wrapped-heading series
- [x] Docs build clean (`docs/make.jl`, no errors) · rebased on `origin/main` after #330
- [x] GitHub CI: 7/7 pass — Julia 1.12 on ubuntu (monolith and kernel), macOS and Windows, Julia 1.13 on ubuntu, Documentation, reuse-lint
- [ ] Local full suite and local REUSE lint: not run — `reuse` is not on this box and `agent ci-local` refuses on this task for the reason in the aside below. CI above is the authority.
- Risk: the test asserts that the panel draws and where its y-axis sits, not what it plots. A log the suite builds carries `steering = 0`, so every `g_k` is `NaN` — and the unit question above is why I did not pin a number to it.

### Scope

+41 / -157 across 4 files, of which `ext/SymbolicAWEModelsMakieExt.jl` is +22 / -154: the `@info` calls, the `window_mean` closure, the commented-out "alpha mystery" and both heading-rate loops out, the helper and the `ylim` key in. Not stacked.

_Aside, not about this repo: `agent intent` refused this turn — the task row carries `worktree` and `branch` as NULL, so `speakingFor()` matches no task, and `ask`, `say`, `reply` and `retract` go with it. The intent note on #329 is the same edit made by hand with `gh api`. Another session appears to be on it already._

Closes #329 · task `SymbolicAWEModels.jl-329`


1-Bort-1 (commented):

**Good**

- The crash is real and exactly as described: `cs_over_us_vec` is referenced at ext/SymbolicAWEModelsMakieExt.jl:1835 on main with no definition anywhere in the repo, reachable only under `plot_gk`.
- `ylim = gk_ylims` uses the key the renderer actually reads — verified the panel loop at ext:2380-2381 tests `haskey(panel, :ylim)`, and no gk panel pushed it before.
- `compute_heading_rate` is a real dedup, not a third copy: `wrap_to_pi` already exists at src/system_structure/transforms.jl:33 as `mod(angle+π,2π)-π`, which reproduces the per-step result of the unwrap loop, and grep shows both loop copies (ext:1430, ext:1795) are the only ones in src/ext.
- The `aoa_ylims` doc removal is correct: the vector method's keyword list (ext:1326-1365) has no `aoa_ylims` and no `kwargs...` catch-all, so the documented call was a MethodError; `ylims=Dict(:aoa=>...)` is served by `get_ylim(:aoa)` at ext:2030.
- CHANGELOG bullet lands under `## Unreleased / ### Fixed` on the branch, one entry, user-visible phrasing — checked the branch file, not the diff hunk.
- +41/-157 with the file shrinking; the deleted block is a debugging session (12 `@info`, a `window_mean` closure, ~90 commented-out lines) and nothing else rides along — diffstat over origin/main shows 4 files, all named in the card.
- New test names the behaviour it protects and reuses the existing `sys1/sys2/lg1/lg2` fixtures and the file's numbered-banner style, so no new `system_name` family is compiled.

**Not good**

- `turn_radius_ylims` has the identical one-letter defect (pushes `ylims` at ext:1566, renderer reads `ylim`) plus a documented default that disagrees with the signature; noted in the card and deferred, and the docs line touched here now advertises both limits as if both work.
- The gk fixture logs `steering = 0`, so every plotted value is `NaN` — the test pins that the panel draws and where its axis sits, never the formula; acknowledged in the card, but the branch stays value-untested.
- `SysState.steering` is read as a percentage here while `update_sys_state!` writes degrees of twist difference and KiteUtils documents `[-1..1]`; correctly left alone as a separate question, but the panel ships computing a number none of the three readings agree on.
- The two `compute_heading_rate` callers still stamp the result on different time samples (`time[1:end-1]` vs `time[2:end]`) and only one clamps 200 °/s jumps; pre-existing and deliberately out of scope, so the shared helper does not yet mean a shared curve.

Issue #337, opened 2026-09-15
Delete outdated plotting code


1-Bart-1:
There is some code that is specific to the V3 kite geometry or ram air kite geometry left. This should be removed both in src and plotting. For example the plot_gk is broken.

1-Bort-1:
Merging origin/main (646b2a10) into #338 to resolve the conflicts, keeping both sides; no other change to the branch.

PR #338, opened 2026-09-15
Delete the V3 kite's steering panels and the src leftovers that fed them

1-Bort-1:
### TL;DR

The multi-panel log plot no longer carries `plot_gk` and `plot_us`: they draw the four-line V3 kite's steering law, which lives in V3Kite.jl, and `plot_gk` has thrown before drawing anything since v0.12.0. Four smaller leftovers of the same split go with them, none of them reachable code, along with the doc lines that described keywords this method does not have.

### What was wrong

`plot_gk` reads `cs_over_us_vec` at what was ext:1836, a name defined nowhere in the repository since `calculate_cs` and the `plot_cs` panel were removed in v0.12.0. On `origin/main` the panel therefore ends the whole call:

```
[ Info: turn-rate 57.29578018188478
[ Info: v_app 20.0
[ Info: us_seg 5.0%
[ Info: gk 57.29578018188478
[ Info: alpha-vsm 0.0
ERROR: UndefVarError: `cs_over_us_vec` not defined in `SymbolicAWEModelsMakieExt`
  @ SymbolicAWEModelsMakieExt ext/SymbolicAWEModelsMakieExt.jl:1836
```

#329 proposed repairing it and #333 did. Closing that one, you said the panel is non-general code left over from before V3Kite was split out and should be deleted instead; this is that.

The reason it is non-general is worth stating, because it decided how far the deletion goes. `g_k` and `u_s` are the gain and the input of the turn-rate law `ψ̇ = g_k · v_a · u_s`, which is how a KCU-actuated four-line kite steers. Both panels read `syslog.steering`, and the only thing that ever wrote it was `update_sys_state!`:

```julia
ss.depower = rad2deg(mean(ss.twist_angles))
ss.steering = rad2deg(ss.twist_angles[outer] - ss.twist_angles[1])
```

Those are station-twist statistics wearing the names of KCU control inputs, in degrees, where `KiteUtils` documents `steering` as `[-1..1]` and `depower` as a percentage — three readings that never agreed. Nothing else in the package reads either field, so with the panels gone the two lines are dead and go too.

### What else went

- `const R_ENU2NED` in the extension — the last trace of the deleted `compute_ekf_yaw_and_rate` helpers, with zero uses anywhere.
- `parse_segment_type` in `src/yaml_loader.jl` — parses the `SegmentType` YAML column, which the loader has rejected outright since it was removed. No caller.
- `scripts/extrapolate_polars.jl` — opens `data/v3/`, deleted in `fdf2ab16`, so it has not run since. Not referenced by `docs/make.jl`, `scripts/build_docu.jl` or CI.
- `src/model_management.jl` still described `:physical_model` as an enumeration of ram-air variants (`ram, simple_ram, 4_attach_ram`); it is a free-form cache-key string.
- The docstring of the multi-panel plot documented `aoa_ylims`, which is not a keyword of that method (a caller passing it gets a `MethodError`), and `suffix::String`, whose real name is `suffixes` and which takes one per log. `turn_radius_ylims` was documented as defaulting to `(0.0, 100.0)` where the signature says `(0.0, 70.0)`.

`docs/logo.jl` stays exactly as it is. It reads the same deleted `data/v3/V3_25.obj`, which is why I had it in the first draft, but it renders the OpenSourceAWE logo and that is a reason to keep the script and find it the mesh, not to delete it. Nothing in this branch touches it now.

### What I left, and would raise

- **`turn_radius_ylims` is dead the same way `gk_ylims` was.** Its panel pushes `ylims =` where the renderer tests `haskey(panel, :ylim)`, so the keyword is accepted and ignored. Fixing the letter would make that panel start clamping its axis, which is a behaviour change to generic code and not this diff's idea. I corrected only the documented default so the docstring stops disagreeing with the signature.
- **`SegmentType` with `POWER_LINE` / `STEERING_LINE` / `BRIDLE` is still exported.** It is V3 line taxonomy, but it is kept deliberately so that old code gets a deprecation error rather than an `UndefVarError`, and removing an exported name is yours to call. I deleted only the parser nothing could reach.
- **`data/2plate_kite/vsm_settings.yaml:45` still names its wing `V3_Kite`.** A label in a generic fixture. I left it because I could not rule out the VSM settings feeding a model cache key, and invalidating the shared cache on the box costs every worktree a rebuild.
- `foil_file: "ram_air_kite/ram_air_kite_foil.dat"` appears in about twenty test YAML blocks and is a grep magnet, but it is inert: `foil_file` is only hashed into the cache key, never opened, and the path does not exist on disk.

Logs written after this carry zeros in the `depower` and `steering` columns instead of twist statistics. That is the one user-visible effect, and it is the half of the change I would look at hardest.

### Verification

- [x] Reproduced first: `UndefVarError: cs_over_us_vec not defined in SymbolicAWEModelsMakieExt` at `ext/SymbolicAWEModelsMakieExt.jl:1836`, quoted above
- [x] After: `plot(sys, lg; plot_default=false, plot_v_app=true)` returns a `Figure`; the same call with `plot_gk=true` is now a `MethodError`, the keyword being gone
- [x] `test/test_makie_extension.jl`, the file that covers `plot`, `record` and `replay`: 21/21 on the head `2c9171a9` (3m16s, juliaserver, GLMakie headless)
- [x] `test_yaml_variables` 18/18 · `test_yaml_bodies` 16/16 · `test_yaml_weighted_ref` 16/16 · `test_beam_replay` 7/7 (the `update_sys_state!` round-trip) — all PASS at `b9ca6660`
- [x] Merged `origin/main` `646b2a10` in at `2c9171a9`. The one conflict was `CHANGELOG.md`, where main's new `### Added` (`set_unstretched_length!`, the split `reinit!` steps) and this branch's `### Fixed` both opened `## Unreleased`; both kept, Added first. Nothing main brought in touches the extension or the deleted code, and no reference to a removed name is left outside `CHANGELOG.md`
- [x] Local full suite on `7a3f06fd`, the head before this merge: PASS, 3557/3557 in 32 min · on `2c9171a9`: started, result posts to this thread
- [ ] GitHub CI on `2c9171a9`: pending. The 1.12 Windows monolith cell that was red on the previous head failed for #345, fixed on main by #348, which this merge brings in
- [ ] Docs build and REUSE lint not run locally — `docs/` has no manifest in the worktree and `reuse` is not installed on this box; CI's `Documentation` and `reuse-lint` jobs are the check
- [ ] `jetls`: **not run** — the binary is not on `PATH` here
- Benchmark: n/a — deletions only, none on a simulation path.
- Risk: someone downstream reads `SysState.steering` off a log this package wrote and now gets zeros. Nothing in this repo or its examples does, and the two panels that did are gone in the same commit.

### Scope

+17 / −287 across 7 files: 195 deleted lines are the extension (the two panels, their three keywords, `R_ENU2NED` and two stale docstring lines), 75 are `scripts/extrapolate_polars.jl`, 13 are the src deletions and 3 the docs page. Of the 17 added lines, 11 are the changelog entry and 6 are corrected doc lines; no logic was written. Not stacked.

Refs #337 · task `SymbolicAWEModels.jl-337`


ext/SymbolicAWEModelsMakieExt.jl
```diff
@@ -1252,20 +1248,18 @@ Create a multi-panel plot of key simulation results from a `SysLog`.
...
 - `plot_distance::Bool=false`: Show the panel with the kite distance from origin (norm of position).
 - `plot_yaw_rate::Bool=false`: Show yaw rate `dψ/dt` derived from the wind-referenced heading.
-- `turn_radius_ylims::Union{Nothing,Tuple}=(0.0, 100.0)`: Y-axis limits for the turn-radius panel (`nothing` leaves autoscaling).
+- `turn_radius_ylims::Union{Nothing,Tuple}=(0.0, 70.0)`: Y-axis limits for the turn-radius panel (`nothing` leaves autoscaling).
 - `plot_cone_angle::Bool=false`: Show the panel with the cone angle (angle between wind vector and normalized kite position).
 - `plot_old_heading::Bool=false`: Show the old heading calculated from orientation quaternion (angle between -R_b_w[:,1] and -R_v_w[:,1]).
 - `plot_winch_force::Bool=plot_default`: Show the panel with the winch forces.
 - `plot_set_values::Bool=false`: Show the panel with the set torque values.
-- `suffix::String=" - " * sys.name`: Suffix to append to plot labels.
+- `suffixes::Vector{String}`: One label suffix per log; defaults to the system names.
```
1-Bort-1, line 1262: MINOR: `suffixes` does not default to the system names for a single log — `actual_suffixes` is `[""]` when `length(logs) == 1`, so the new docstring (and the same sentence at docs/src/exported_functions.md:167) is wrong in the common case. It replaces one inaccurate doc line with another.

ext/SymbolicAWEModelsMakieExt.jl
```diff
@@ -1252,20 +1248,18 @@ Create a multi-panel plot of key simulation results from a `SysLog`.
...
-- `aoa_ylims::Union{Nothing,Tuple}=nothing`: Y-axis limits for the AoA panel (`nothing` leaves autoscaling).
 - `plot_heading::Bool=plot_default`: Show the panel with the kite's heading and course angles.
 - `plot_kiteutils_course::Bool=false`: Also plot course calculated using KiteUtils.calc_course.
-- `gk_ylims::Union{Nothing,Tuple}=(0.0, 10.0)`: Y-axis limits for the gk panel (`nothing` leaves autoscaling).
 - `plot_elevation::Bool=false`: Show the panel with the kite's elevation angle.
 - `plot_azimuth::Bool=false`: Show the panel with the kite's azimuth angle.
 - `plot_distance::Bool=false`: Show the panel with the kite distance from origin (norm of position).
 - `plot_yaw_rate::Bool=false`: Show yaw rate `dψ/dt` derived from the wind-referenced heading.
-- `turn_radius_ylims::Union{Nothing,Tuple}=(0.0, 100.0)`: Y-axis limits for the turn-radius panel (`nothing` leaves autoscaling).
+- `turn_radius_ylims::Union{Nothing,Tuple}=(0.0, 70.0)`: Y-axis limits for the turn-radius panel (`nothing` leaves autoscaling).
```
1-Bort-1, line 1257: MINOR: This line still presents `turn_radius_ylims` as setting the panel's limits, while the PR card establishes it is accepted and ignored (panel pushes `ylims =`, renderer tests `:ylim`). Correcting only the default leaves the docstring promising behaviour that does not happen; saying it is currently ignored costs one clause and no behaviour change.

1-Bort-1 (commented):

**Good**

- Deletion is complete: `rg` for plot_gk/plot_us/gk_ylims/cs_over_us_vec across src, ext, test, examples and docs returns only CHANGELOG and untracked .agent files.
- Removal takes its feeders with it (§2): the `ss.depower`/`ss.steering` writes were the panels' only source and nothing else in the repo reads either field.
- No orphaned imports: `Statistics` (mean) and `StaticArrays` (SMatrix) both still have live uses in src and ext after `R_ENU2NED` and `window_mean` went.
- `docs/logo.jl` and `scripts/extrapolate_polars.jl` are referenced nowhere — not by `docs/make.jl`, `scripts/build_docu.jl`, CI or README — and both open the deleted `data/v3/`.
- `:physical_model` docstring correction checked against the code: it is a free-form string from `set.physical_model` hashed into the cache filename, not the enumeration the old line claimed.
- `turn_radius_ylims=(0.0, 70.0)` in the corrected docstring now matches the signature at ext:1325.
- Doc and changelog surfaces updated in the same diff; entry is marked BREAKING and matches the file's existing house style; tree clean, no plan or scratch files committed.
- Scope discipline: the `SegmentType` enum kept (removing an exported name is the maintainer's call), only the unreachable parser deleted.

**Not good**

- `ext/SymbolicAWEModelsMakieExt.jl:1262` — `suffixes` does not default to the system names for a single log — `actual_suffixes` is `[""]` when `length(logs) == 1`, so the new docstring (and the same sentence at docs/src/exported_functions.md:167) is wrong in the common case. It replaces one inaccurate doc line with another.
- `ext/SymbolicAWEModelsMakieExt.jl:1257` — This line still presents `turn_radius_ylims` as setting the panel's limits, while the PR card establishes it is accepted and ignored (panel pushes `ylims =`, renderer tests `:ylim`). Correcting only the default leaves the docstring promising behaviour that does not happen; saying it is currently ignored costs one clause and no behaviour change.
- `SysState.depower`/`steering` now silently log 0.0 rather than erroring — a downstream log reader sees plausible data; the card flags the risk, but a correctly-named twist column would have been the other option.
- `docs/src/assets/logo.png` is still committed while its generator is deleted, so the logo is now unregenerable — acceptable only because its input data is gone too.
- `- :physical_model: ... (affects geometry)` is a slight overstatement: it is a label in the cache key, not itself a geometry input.

docs/logo.jl
1-Bart-1, line 1: Don't delete this, we use this as the OpenSourceAWE logo

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+### Removed
+- BREAKING: the `plot_gk` and `plot_us` panels of the multi-panel log plot and
```
1-Bart-1, line 6: Don't report as breaking, this is a fix bc it was already broken and now just removed.

1-Bart-1 (changes requested):

docs/logo.jl
1-Bort-1: Restored in fa67ab88, unchanged.

CHANGELOG.md
1-Bort-1: Rewritten in fa67ab88 as a Fixed entry, no BREAKING marker.

1-Bort-1:
**Local full suite: PASS** (38 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `a14820503`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34958352560/job/104345728387

1-Bart-1 (approved):

1-Bart-1:
Resolve the conflicts

1-Bort-1:
Resolved in 2c9171a9: merged main 646b2a10; the only conflict was CHANGELOG.md, both Unreleased entries kept.

PR #339, opened 2026-09-15
Colour the replayed segments by spring force, behind a "Spring force" checkbox

### LINKED ISSUE #336: Add toggle in replay to show or hide spring force coloring
Color segments by spring force, add toggle for this to show or hide, make sure the spring force is available and copied from log when updating the sys struct from the log such that this works in replay.

1-Bort-1:
### TL;DR

The viewer could already colour segments by spring force, but only as a construction-time `force_color=true`, and in replay every segment read zero — `update_from_sysstate!` never wrote `sys.segments`, so the colouring had nothing to ramp over. The restore is added and the flag becomes a replay checkbox.

### What was wrong

`update_sys_state!` has been writing `ss.spring_force[segment.idx] = segment.force` for every segment all along, and the column survives Arrow, so the log was never the problem. `update_from_sysstate!` simply did not read it back: it ended on a comment saying segment forces come from the symbolic getters, which is true while a simulation is running and false for a structure that only ever gets fed from a log. A 12-segment tether logged at 39.8 N down to 5.9 N came back as twelve zeros, `calculate_segment_force_colors` saw a zero range, and every segment fell back to the plain grey.

The second half is that `force_color` was only reachable by passing it to `plot` before the window opened. There was no way to turn the colouring on or off while watching a replay, which is what the issue asks for.

### What changed

`update_from_sysstate!` now restores `segment.force` from `sys_state.spring_force`, guarded on the column's length the same way `pulley_len` and `twist_vel` already are; a log written before the column existed leaves the force `NaN` rather than a plausible-looking zero.

The replay checkboxes gain a "Spring force" entry. The checkbox tuple's second element has always been the `PLOT_LAYERS` key the box acts on, and it still is — `:segment_colors_obs` is the key holding the segment colour observable, so the existing `haskey` filter drops the box for a system with no segments without any special case. What the key *means* is decided in one place, `apply_view_toggle!`, which recolours for that key and shows or hides the layer for every other one.

`force_color` is `replay`'s own keyword on both methods, passed to `plot` and used as the box's initial state, so there is one name for the setting rather than a new `show_force_color` alias beside it. It stays off by default, which is what every existing `force_color` default is; the box is how you turn it on.

### Riding along, because the diff already opened them

- The `force_color ? ramp : fill` branch existed three times — at construction, in `update_plot_observables!` and in the hover handler. It is now `segment_display_colors`, which also drops the NaN hazard the copies had: `minimum`/`maximum` over raw forces made one unknown force blank the whole layer, exactly what `finite_force_arrows` ten lines above documents guarding against. It now ramps over the finite forces and leaves the rest on the base colour.
- The hover handler read a `force_color` captured at construction, so with a runtime toggle the first hover after a click would have undone it. It reads `PLOT_FORCE_COLOR[]` now, and `setup_segment_hover_events!` loses the kwarg no caller still needs.
- The toggle list was byte-identical in both `replay` methods; it is one `replay_toggles`, and its entries are named `(label, key, enabled)` rather than positional tuples indexed as `t[2]`.
- The `update_from_sysstate!` docstring claimed aerodynamic forces and moments were NaN'd, which was never true — they are restored. It now lists what the function actually NaNs.

### Where I would push back

- The ramp is min–max **per frame**, so a segment's colour is its rank within that frame, not an absolute force, and the scale slides as you scrub. That is pre-existing and I left it alone, but a fixed range over the whole log would be the more useful thing and is a small follow-up.
- `plot_with_panes` still passes `force_color` down for the *initial* colours while the hover handler it installs reads the global. They cannot disagree today, because the only path that installs hover is `MakieControlPlots.plot`, which sets the global immediately before. It is an invariant held by call order rather than by construction.
- `plot_body_frame` (`ext/SymbolicAWEModelsMakieExt.jl` around 4145 and 4258) holds two more copies of the same min–max → red/green ramp, for point depth rather than force, duplicated between themselves as well. This PR had no reason to open that function, so they are left for a `cleanup:` PR.
- `segment.len` is still not restored during replay. Nothing draws it, so nothing is visibly wrong, but it is the same class of gap as the one this fixes.

### Figures

before: a replayed 12-segment tether, every segment the same grey because the restored spring forces are all zero

after: the same frame with the Spring force box ticked -- red at the anchor carrying 39.8 N, green at the free end carrying 5.9 N

Same model, same frame, same camera; the before was rendered with the restore taken back out of the working tree, so the only difference is the fix.

The box itself, in the legend at the top right, where the other layer checkboxes live:

the replay legend at the top right, with the Spring force box ticked; this fixture has no wings, beams or panels, so the other checkboxes are filtered out

### Verification

- [x] Reproduced first: log holds `spring_force = 17.5758 N`, `update_from_sysstate!` leaves `segment.force = 0.0`; on the 12-segment tether, `[39.81, 36.69, … 5.9]` logged against `[0.0 × 12]` restored
- [x] `test/test_segment.jl` "spring force survives the SysLog round trip" red before (`[0.0] ≈ [17.57577071651983] (rtol=1.0e-5)`), green after — 91/91
- [x] `test/test_makie_extension.jl` "Spring force checkbox recolours the replayed segments" red before (2 × `!allequal(colors[])`), green after — 21/21. It replays into `sys2`, which never ran the simulation, so the only forces it can colour by are the log's
- [x] Local full suite: PASS (3557/3557, 42m37s) · GitHub CI: triggered by this pull request, so it had not run when this was written
- [x] Docs build clean (only the pre-existing size warnings of #301) · up to date with `main` (rebased onto c45a59ab)
- [ ] REUSE lint: **not run locally** — `reuse` is not installed on this box; CI's `reuse-lint` job covers it. No files were added, so no new SPDX headers are involved.
- Benchmark: n/a — nothing on the ODE path changed.
- Risk: the restore writes `NaN` for a log without the column, and `NaN` reaching a field that seeds `u0` as a torn variable would be worse than the zero it replaces. The full suite's `validate_sysstate_roundtrip` runs this path for every model it builds and stayed green, and segment force is not in `state_vars`, but that is the line to look at first.

### Scope

+175 / −81 across 5 files. The extension is +95/−74 for a whole feature because three copies of the colour branch, two copies of the toggle list and two copies of the checkbox-application became one each. `test/test_makie_extension.jl` grows its fixture from 2 points / 1 segment to 3 / 2, since one segment can never make a colour ramp.

Closes #336 · task `SymbolicAWEModels.jl-336`


ext/SymbolicAWEModelsMakieExt.jl
```diff
@@ -3526,7 +3556,7 @@ This is a reusable function that can be used by both single and multi-system rep
 """
 function setup_replay_controls!(scene, n_frames, update_frame!, get_time, get_dt;
                                  replay_speed=1.0, autoplay=false, loop=false,
-                                 toggles=Tuple{String, Symbol, Bool}[])
+                                 toggles=[])
```
1-Bort-1, line 3559: MINOR: `toggles` changed from `Tuple{String,Symbol,Bool}[]` to an untyped `[]` whose elements must now be named tuples with `.label`/`.key`/`.enabled`, but the docstring's signature line and Keyword Arguments section still do not mention `toggles` at all. A caller passing the old positional tuples fails deep inside the loop on `toggle.enabled` with nothing in the docs to point at the contract.

1-Bort-1 (commented):

**Good**

- The restore matches the log's layout: `ss.spring_force[segment.idx]` in `src/symbolic_awe_model.jl:411` writes what `update_from_sysstate!` now reads, and KiteUtils declares `spring_force::MVector{S}` as one per segment, so the length guard is the right one.
- `segment_display_colors` genuinely unifies the three copies of the `force_color ? ramp : fill` branch — `rg calculate_segment_force_colors` now returns nothing repo-wide, and the construction, per-frame and hover paths all call the one function.
- The NaN guard inside the ramp is a real fix, not a restatement: `known` filters non-finite forces so one unknown segment no longer collapses the shared min–max scale, which is the hazard `finite_force_arrows` already documents.
- `segment.force` feeds only logging and `spring_forces()` (checked all `.force` uses under `src/`), so the new NaN-on-missing-column path cannot leak into `u0` — the card's stated risk is bounded.
- Tests match the claims: `test_segment.jl` covers the save/load round trip through Arrow, and `test_makie_extension.jl` drives `apply_view_toggle!` on/off/on against a structure that never simulated, so the colours can only come from the log.
- Both `replay` methods reuse `system_name` families and one `replay_toggles`, and the `(label, key, enabled)` named tuple removes the `t[2]` positional indexing that made the old filter opaque.

**Not good**

- `ext/SymbolicAWEModelsMakieExt.jl:3559` — `toggles` changed from `Tuple{String,Symbol,Bool}[]` to an untyped `[]` whose elements must now be named tuples with `.label`/`.key`/`.enabled`, but the docstring's signature line and Keyword Arguments section still do not mention `toggles` at all. A caller passing the old positional tuples fails deep inside the loop on `toggle.enabled` with nothing in the docs to point at the contract.
- `has_spring_force` guards on total length but indexes by `segment.idx`, while the pulley/station loops beside it guard per-element with `idx <= length(...)`; equivalent today, inconsistent to read.
- A log whose segment count disagrees with the structure silently NaNs every segment — indistinguishable from a pre-column log, with no warning at the one place that could tell.
- The multi-system replay shows the Spring force box but only the primary system recolours; the ghosts stay grey, which is consistent but unstated anywhere.
- The test fixture header explains *why* two segments are needed ("one carries the lower mass and the other carries both") — that rationale belongs in the PR, not the file (§3).
- Three inline comments in the new makie testset (`# plain segment colour by default`, `# the two segments differ in tension`) restate the assertions above them.
- The new `test_segment.jl` case builds a model and steps five times to obtain forces a hand-built `SysState` would have supplied; justified only by also covering the Arrow column, which the test does not say.

PR #340, opened 2026-09-15
Let the CI test matrix stop at the first failing job

### LINKED ISSUE #307: cleanup: CI's test matrix carries fail-fast: false

1-Bort-1:
### TL;DR

`.github/workflows/CI.yml` loses `fail-fast: false` from the `test-with-code-coverage` matrix, so a failing cell now cancels its siblings instead of letting them re-prove the same failure for another hour. AGENTS.md §4.10 asks for exactly this, and the line was never a decision anyone made. Two dead things in the same job go with it: a three-line `exclude` that has excluded nothing since 2026-04, and a comment claiming a cell runtime 50% over what the cell measures — the claim this issue was written from. Five deletions, one file, nothing added.

### It was scaffolding, not a choice

The issue asks to check first whether someone wanted the windows cell to keep reporting when ubuntu fails. They did not. `git log -S 'fail-fast' -- .github/workflows/` gives three commits: `6f9ec987` ("Files generated by PkgTemplates", 2025-06-28) introduced the line as the template's default, and `438ff46a` and `f6b61037` carried it forward while restructuring the jobs without touching it. Every `strategy:` block in this repo that somebody wrote by hand — `f45e39fb`, `55bc4ec8` — says `fail-fast: true`.

### What it saves, and where it saves nothing

Three runs on this branch have failed under the change, and they bracket the range.

Run `35008599639` on `ef446976`, the head, is the clearest. The windows cell failed at 19:25:12, 46m52s in, and the three cells still running were cancelled within 98 seconds — ubuntu kernel at 47m45s elapsed, macOS at 48m05s, ubuntu monolith at 48m30s. The two ubuntu cells had taken 72m11s and 67m35s to complete on run `34998044576`, one comment earlier on the same branch, so **24m26s and 19m05s of runner time went unspent**. The macOS cell was already past the 37m23s and 39m57s it takes when it is well, so nothing is claimed for it.

Run `34976698873` on `7b83453e` is the same shape: windows failed at 15:48:52 after 91m11s, the two ubuntu cells were cancelled seconds later at 45m34s and 41m56s elapsed, against 67m17s / 66m59s / 68m32s and 68m22s / 85m54s / 64m52s to completion on `main` (runs `34957804412`, `34944466963`, `34876174462`) — roughly 21 and 25 minutes unspent.

Run `34998044576` on `ba546100` is the other end: the macOS cell failed at 92m52s and windows was cancelled 2m08s later at 95m07s elapsed, past the 71m28s that same cell had taken to finish green on `393c0ea0`. It gave back nothing.

Cells here run from 19m29s to 95m07s, so fail-fast pays when a fast cell fails first and close to nothing when the slowest one does. Two of the three runs were the good case, the third cost nothing, and none of them can cost more than the old behaviour did.

### The answer you lose, and what it takes to get it back

This is the real cost of the change, and it is sharper than "you have to re-run something". The issue put it as "the two do fail for different reasons often enough that someone may have wanted both answers", and that is right.

**Re-running the cell that failed works.** On run `34998044576` the macOS cell was the failure, and `gh run rerun <run-id> --job <job-id>` restarted it alone at 18:32:32 with every other conclusion carried forward from attempt 1.

**Re-running a cell that was cancelled does not, while the matrix still holds a failure.** Tried on this head — `gh run rerun 35008599639 --job 104514917268` for the cancelled macOS cell — and GitHub cancelled it one second after it started, 19:28:26 to 19:28:27, annotating it:

> The strategy configuration was canceled because "test-with-code-coverage.\_1\_12\_windows-latest\_x64\_" failed

That is `fail-fast` applying to the re-run: the new job joins a matrix that already contains a failure, so the strategy cancels it before it does anything. So while windows stays red for #345, **a cancelled cell on this head cannot be made to report on its own**: the only way through is to clear the failure first, by re-running the red cell until it passes or by fixing what it is red for. Re-running the whole set puts the red cell back in the matrix, where it fails again 47 minutes in and cancels whichever siblings are still going.

The honest statement of the trade is therefore: the price of the change is that a matrix with one persistently broken cell reports nothing but that cell, until it is fixed. What makes it affordable here is that `agent ci-local` runs the whole suite on the box independently of GitHub — three PASSes on this branch, below — and that a cell which stays red is meant to be fixed rather than run around. The alternative, `fail-fast: false`, buys the second answer by paying for every cell on every failing run, which is what AGENTS.md §4.10 decided against.

### Codecov does not lose a report

Worth checking, because on run `34976698873` both cancelled cells were the `coverage: true` ones. It costs nothing: `julia-processcoverage` and `codecov-action` are guarded by `!cancelled()` (`CI.yml:133,136`) and were `skipped` in both, so Codecov was never going to receive anything from a run whose coverage cell died. Both coverage cells completed and uploaded on run `34998044576`.

### The `exclude` in the same block excluded nothing

Found while reading the file around the deletion, and it rides here because it is three dead lines in the very `strategy:` block this diff already opens (CLEAN_CODE §1).

`exclude: os: macOS-latest, arch: x64` was live in `a117a539`, when `os` listed `ubuntu-latest`, `windows-latest` and `macOS-latest`: it took the macOS-x64 combination back out of the cross-product so that only the aarch64 `include` survived. `3fe6d70a` dropped `macOS-latest` from `os` and left the `exclude` behind. Since then the cross-product has had no macOS row for it to match, and the one macOS job comes from `include`, which GitHub processes after `exclude` and never filters. GitHub's own matrix expansion confirms it: run `34976698873` with the `exclude` and run `34998044576` without it produce the same five cells — ubuntu monolith, ubuntu kernel, ubuntu 1.13, windows monolith, macOS aarch64. There was never a macOS-x64 job for it to remove.

### The comment above `timeout-minutes` is where this issue's "~100 minutes" came from

Two lines above the `fail-fast: false` deletion, `CI.yml` said:

```yaml
    # The kernel backend needs ~100 min of the budget and grows with the suite.
    timeout-minutes: 180
```

The kernel cell does not need ~100 minutes. It takes 67m17s, 66m59s and 68m32s on `main` (runs `34957804412`, `34944466963`, `34876174462`) and 72m11s on this branch. The comment arrived with the budget itself — `03ee87d6`, then carried through the `cb904f2e` / `e70a0d8d` revert pair — and its number has not been true since whatever suite it was measured against.

It is worth fixing rather than noting, because it is not an inert stale comment. The issue that opened this pull request reads "the kernel-backend cell alone is budgeted at ~100 minutes": this line, quoted as if it were a measurement. A comment tracking a value in a config file is the exact shape CLEAN_CODE §3 rules out, for this reason — the value moves and the comment does not. So the line goes, `timeout-minutes: 180` speaks for itself as a cap, and the measurement lives here in the pull request where it cannot rot back into the file. Behaviour-preserving, in the job the diff already opens, so AGENTS.md §4 routes it here.

### Reading the checks on the head

Run `35008599639` on `ef446976`:

| cell | conclusion | elapsed |
| --- | --- | --- |
| `reuse-lint` | success | 14s |
| `Documentation` | success | 9m04s |
| Julia 1.13 · ubuntu · monolith | success | 19m29s |
| Julia 1.12 · windows · x64 · monolith | **failure** | 46m52s |
| Julia 1.12 · ubuntu · kernel | cancelled | 47m45s |
| Julia 1.12 · macOS · aarch64 · monolith | cancelled | 48m05s |
| Julia 1.12 · ubuntu · x64 · monolith | cancelled | 48m30s |

The one `failure` is #345, not this diff: `test_segment.jl:163`, `Got exception outside of a @test`, `IOError: rm("C:\Users\RUNNER~1\AppData\Local\Temp\jl_g7HfHG"): directory not empty (ENOTEMPTY)`, and the suite otherwise `3536 passed, 0 failed, 1 errored, 1 broken`. It arrived on `main` with #339 and open PR #348 deletes the exact `rm(tmpdir; recursive=true)` that throws, inside that same `Segment Tests` testset. AGENTS.md §7: red on `main` before this branch started is not this branch's bug. Not stacked on #348 either — that touches `test/test_segment.jl`, this touches `.github/workflows/CI.yml`, no shared code and no ordering between them. **Windows cannot go green here until #348 lands**, and every other cell either passed or was cancelled by this change.

The three `cancelled` cells are the change working. They do make the run read worse than it is: `gh pr checks` prints `cancelled` as `fail`, so this shows as four red cells when one thing broke. On a red run, read `.jobs[].conclusion` from the jobs API first and find the single `failure`.

### The red the thread asked about was a lost runner

The previous head, `ba546100`, went red on the macOS cell (run `34998044576`, job `104479170659`) and the thread reported that its log could not be read. **No test failed there — the runner died.** The failure annotation is GitHub's, not Julia's: "The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error." The jobs API agrees: step 9, `julia-actions/julia-runtest`, still reads `in_progress` with no conclusion and steps 10 to 12 never started, so the job died inside the step and nothing reported. (No log of a run is readable while any cell of it is alive, which is the "could not be read"; `check-runs/<job>/annotations` answers regardless.)

It could not have been the diff. `7b83453e` to `ba546100` is the three-line `exclude` deletion and nothing else, and GitHub expanded the identical five cells either side of it. The same cell passed the same suite twice on this branch — 37m23s on `7b83453e`, 39m57s on `393c0ea0` — and takes 49m46s / 47m48s / 87m03s on `main`. It ran 92m52s before dropping off, past the slowest of those, which is a starved runner rather than a slow test. Re-running that cell alone on the current head to prove it would have settled it, but fail-fast cancels a re-run of a cancelled cell while windows is red — see above.

Everything else on that head was green: `reuse-lint`, `Documentation` and all three ubuntu cells (`61m50s`, `67m35s`, `72m11s`), with windows cancelled by this change at 95m07s.

### The local suite's one failure is #305

`agent ci-local` on `7b83453e` came back `FAIL in 22m, exit 1` with 1309 passed and 1 failed. It is worth being precise about what that covers: the mirror runs `Pkg.test()` under failfast, `test_helpers.jl` sorts fifteenth of the forty-nine files the suite selects on the monolith backend, and the run stopped there — so thirty-four files never ran. The one failure is `test_helpers.jl:50`, the assertion that the tracked `.default` manifests are no older than `Project.toml`:

```
test_helpers.jl:50  mtime("Manifest-v1.12.toml.default") >= project_mtime - 1.0
  Evaluated: 1.7894626410955336e9 >= 1.7894890907210274e9
```

`git diff origin/main -- Project.toml 'Manifest-*.toml.default'` is empty — the files are byte-identical to `main`, and only their mtimes differ, because something in the worktree's session rewrote `Project.toml` with the same bytes. `touch` on the two `.default`s, tree still clean, same test file: 9 passed, 9 total. Red to green with no content changed, so the assertion is measuring mtime and nothing else. That is #305 exactly, already open and waiting on a human with a replacement check proposed.

Re-run end to end with the mtimes touched, the mirror goes the distance every time: twice on `ba546100` and once on `ef446976`, **PASS in 31m, 27m and 27m, exit 0 each time, all 49 files, 3557 passed and 0 failed each time** (`Testing SymbolicAWEModels... | 3557 3557` at 30m56.5s, 26m46.5s and 26m23.7s). So the branch is clean locally and the `FAIL` above was the proxy and nothing else. Neither is a fix for #305, and a `.github/workflows/CI.yml` diff could not reach a Julia test in any case.

### The local half of §4.10 is already there

Worth recording, because the rule this pull request serves asks for fail-fast "locally and in CI" and `test/runtests.jl:74` runs `@testset verbose = true` with no `failfast`. It needs none: `agent ci-local` runs `Pkg.test()` with `JULIA_TEST_FAILFAST=true`, which Julia's `Test` applies to the outermost testset, and the `7b83453e` run stopping dead at file 15 is that working. Nothing to open here.

### Verification

- [x] Reproduced first: n/a — a workflow line, nothing to reproduce locally. The demonstration is the three runs under "What it saves" above.
- [x] Fail-fast confirmed working on all three failing runs, from the jobs API: run `35008599639`, windows `failure` 19:25:12 then three cells `cancelled` 19:26:05, 19:26:30, 19:26:50; run `34976698873`, windows `failure` 15:48:52 then two `cancelled` 15:49:40 and 15:50:27; run `34998044576`, macOS `failure` 18:28:39 then windows `cancelled` 18:30:47.
- [x] GitHub CI on `393c0ea0`, before `main` was merged in (run `34953724658`): all seven checks success.
- [x] GitHub CI on the head `ef446976` (run `35008599639`): `reuse-lint`, `Documentation` and the 1.13 ubuntu cell success; windows `failure` for #345; the other three `cancelled` by this change. Table above.
- [x] The head's one `failure` is #345, read from the job log: `test_segment.jl:163`, `Got exception outside of a @test`, `IOError: rm(...): directory not empty (ENOTEMPTY)`, `3536 passed, 0 failed, 1 errored, 1 broken`. Open PR #348 deletes the `rm` that throws.
- [x] The `ba546100` macOS failure is infrastructure, not a test: annotation "The hosted runner lost communication with the server", `julia-runtest` still `in_progress` with no conclusion, later steps never started, and the identical tree passed that cell twice on this branch. The cell has not since been re-run green: on the current head it is `cancelled`, and a re-run of it is itself cancelled by the matrix while #345 stands.
- [x] `exclude` deletion is a no-op: run `34976698873` before it and run `34998044576` after it expand to the same five matrix cells, read from the jobs API — ubuntu monolith, ubuntu kernel, ubuntu 1.13, windows monolith, macOS aarch64.
- [x] `CI.yml` parses after all three deletions (`YAML.load_file` in the worktree's session): `strategy` holds only `matrix`, whose keys are `backend`, `arch`, `version`, `os`, `include`; `timeout-minutes` is still 180.
- [x] The deleted runtime claim was wrong: kernel cell 67m17s / 66m59s / 68m32s on `main` and 72m11s on this branch, against "~100 min".
- [x] `test/test_helpers.jl` alone: 7 pass / 2 fail before touching the `.default` mtimes, 9 pass / 9 after, manifests byte-identical to `main` throughout — so that failure is #305's mtime proxy and nothing else.
- [x] Local full suite, mtimes touched: twice on `ba546100` and once on the head `ef446976`, **PASS in 31m, 27m and 27m, exit 0 each time** — all 49 files (`--> test_*` lines in the log), `3557 3557` each time, no `Error During Test`. (The earlier `7b83453e` run stopped at `test_helpers.jl`, file 15 of 49, leaving 34 files unrun.)
- [x] Up to date with `main` — `git rev-list --count HEAD..origin/main` is 0.
- [x] REUSE lint green in CI on every head of this branch, `ef446976` included. Not runnable on this box — `reuse` is not installed and `./bin/reuse_lint` says so.
- [x] Changelog: nothing to add. Five deleted lines of CI configuration change nothing a user of the package can observe.
- [x] Docs: nothing to add. No public symbol added, renamed or removed; the `Documentation` job is green on the head.
- Benchmark: n/a
- Risk: a red matrix now shows cancelled siblings that `gh pr checks` prints as `fail`, so a reader can mistake one failure for several. The jobs API distinguishes them.

### Scope

-5 lines in `.github/workflows/CI.yml`, one file, no other file on the branch, nothing added. `fail-fast: false` is the idea; the three-line `exclude` and the one-line runtime claim are dead lines in the same job the diff opens, each called out above. Not stacked.

Closes #307 · task `SymbolicAWEModels.jl-307`


1-Bort-1 (commented):

**Good**

- Diff is exactly the one line the card promises — `git diff origin/main --stat` is 1 file, 1 deletion, tree clean.
- History claim verified: `438ff46a` copied `fail-fast: false` from the PkgTemplates scaffold, while the hand-written blocks in `f45e39fb`/`55bc4ec8` both chose `fail-fast: true`.
- Deleting the line rather than writing `fail-fast: true` is right — `true` is the GitHub default, so the explicit line would be config stating nothing.
- `rg fail-fast` confirmed clean across `.github/`; `setup-test.yml` and `TagBot.yml` define no matrix, and no workflow keys off CI job names via `workflow_run`/`check_run`.
- Card is honest about what it left out (the dead `exclude` at CI.yml:50-52, `test/runtests.jl` failfast) instead of silently riding them in.

**Not good**

- The comment at CI.yml:134-135 justifies the `!cancelled()` coverage guard by "the backend reading as 0% against the other jobs" — fail-fast now produces exactly that for the *cancelled* sibling coverage job, so the comment is half-true where it stands; unnumbered in this diff, and the card does name the Codecov consequence.
- The dead `exclude:` block is three lines away from the changed line and provably a no-op; leaving it for its own `cleanup:` is defensible but a reviewer opening this file will trip over it first.
- A ~40-line card for a one-line deletion is heavy, though the issue explicitly asked for the was-it-on-purpose evidence.

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `7b83453e7`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34976698873/job/104406115781

1-Bort-1:
**Local full suite: FAIL** (22 min, Julia 1.12.7, one cell of the matrix)

```
  Deserialization Tests             |   10           10     12.7s
  point flap geometry               |    7            7      0.4s
  flap + continuous pressure aero   |   18           18   1m21.4s
  point flap drives the polars      |   10           10   1m41.6s
  Flap δ + beam-anchored point      |   28           28     22.9s
  Flow curvature                    |   76           76     58.3s
  Getter allocations                |    6            6     17.4s
  Getter allocations with a wing    |    4            4     29.3s
  Heading - Circular Path           |   72           72      0.3s
  Heading - Horizontal Circle       |   48           48      0.1s
  Heading - Special Cases           |   22           22      0.2s
  Heading - Near-Vertical Placement |   10           10      0.8s
  Testing helper functions...       |    7     1      8      1.4s
RNG of the outermost testset: Random.Xoshiro(0xcb860e079ae26570, 0x8f3e1feee769e58e, 0xbcd6e8455c881ee1, 0xc57e172ad1d1d829, 0x8fe5995adc85999e)
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/Types.jl:68
    @ Pkg.Operations ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/Operations.jl:2642
  [3] test
    @ ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/Operations.jl:2487 [inlined]
    @ Pkg.API ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:552
  [5] test(pkgs::Vector{PackageSpec}; io::IOContext{IO}, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:169
  [6] test(pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:158
  [7] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::PackageMode, subdir::Nothing, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:186
  [8] test()
    @ Pkg.API ~/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:175
  [9] top-level scope
    @ none:1
 [10] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [11] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:283
 [12] _start()
    @ Base ./client.jl:550
```

1-Bort-1:
**Local full suite: PASS** (26 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - macOS-latest - aarch64 - monolith - pull_request` failed** on `ba546100b`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34998044576/job/104479170659

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `ef4469767`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/35008599639/job/104532184084

PR #342, opened 2026-09-15
Place the rotation a transform's turn_rate names

### LINKED ISSUE #115: Add turn_rate to transform

1-Bort-1:
### TL;DR

`Transform.turn_rate` was parsed from the YAML, converted from degrees and then thrown away: `reinit!` warned that the field was ignored and placed the structure at rest. It is now the third term of `spherical_spin`, so a structure comes out of `reinit!` or `reposition!` already turning about the radial axis through the transform's base, at the rate its own heading advances at.

### What the field meant and what it did

Everything around `turn_rate` was already in place — the field with its `[rad/s]` docstring, the constructor keyword, the YAML parse that reads it in degrees per second, the info line that echoes it back. Only the dynamics were missing, and `reinit!` said so in a `@warn` rather than in the docs, so the setting looked live and behaved as zero.

`spherical_spin` builds the angular velocity of the rigid rotation about the transform's base. It carried the two tangential terms, `elevation_vel` about `-y` of the tangential frame and `azimuth_vel` about world `-z`. The missing one is radial: `apply_heading!` rotates the structure by `delta_heading` about the unit vector from base to placed object, so a heading that advances at `turn_rate` is an angular velocity of `turn_rate` about that same axis. Adding `turn_rate * radial_direction(transform)` is the whole behaviour change.

That radial direction was already being computed, inline and unnamed, by `apply_azimuth_elevation!` to decide where to put the object. Rather than write the expression a second time I lifted it into `radial_direction`, which both callers now use — the one place where the two are the same vector, and the reason the velocity lands on the right axis in the first place.

### How I know the axis is right

The test is a finite difference, so it does not restate the formula. It places the structure at `heading = h₀` with `turn_rate = ω` and keeps the velocities `reinit!` assigns, then places it again at `heading = h₀ + ω·dt` with `turn_rate = 0` and compares: `(pos₁ − pos₀)/dt ≈ vel₀` for every point and every body. The only thing that satisfies it is a rotation about the axis heading actually turns about, at the rate heading actually advances at. Against unchanged code the left side is zero and the right side is not.

### Where I would push back

`turn_rate` is a placement rate, not a sustained one. It sets the velocity state the solver starts from; nothing holds the structure turning afterwards, and the aerodynamics and tether will do what they do. That matches the field's docstring and what `elevation_vel` and `azimuth_vel` already meant, but anyone expecting a commanded turn will not find one here.

### The red ubuntu check is #316

On `c317fe5f` five of six test jobs were green, windows included now that #348 is in. `Julia 1.12 - ubuntu-latest - x64 - monolith` errored once, `VSM solve did not converge in 5000 iterations` inside the steering ramp at `test/test_twist_alignment.jl:123` — the ramp that stops next to the VSM divergence, which #316 owns and which has gone red in that cell on `main` itself (`fa29f14c`). Its static pass printed twist `-0.13351 / -0.00083 / -0.13167`, the values #316 records for the failing runs on other branches, and nothing in `data/2plate_kite` sets a `turn_rate`, so the radial term this branch adds is zero there. The same file passes on this box on the merged tree. I have not shortened the ramp here: that is #316's change and wants a full matrix to confirm.

### Verification

- [x] Reproduced first: `reinit!` warned `turn_rate ... is not zero, but turn_rate dynamics are not yet implemented`, and every point's `vel_w` came back at rest
- [x] `test/test_transform.jl` red before the change (8 of 16 finite-difference assertions per dynamics config), green after; 357/357 on the merged tree `b6b96240`, juliaserver, exit 0
- [x] `test/test_tether_init.jl` 58/58 · `test/test_twist_alignment.jl` 13/13 on `b6b96240` (main's `reinit!` split lands beside this change); `test/test_wing.jl` 171/171 · `test/test_aero_modes.jl` 809/809 on the previous merge
- [ ] Local full suite on `b6b96240`: running
- [ ] GitHub CI on `b6b96240`: running. On `c317fe5f`: all green but ubuntu 1.12 monolith, #316 as above
- [x] Docs build and REUSE lint green on CI for `c317fe5f`; docs not built locally · up to date with `main`
- Benchmark: n/a
- Risk: the sign of the radial term. It is fixed by `apply_heading!`'s own rotation axis and the finite-difference test would catch a flip, but a reader checking by hand should start there.

### Scope

+63 / -15 across 4 files. `src/system_structure/transforms.jl` is the whole behaviour change: three lines in `spherical_spin`, the `radial_direction` extraction that `apply_azimuth_elevation!` now shares, and the removal of the obsolete warning. The rest is the test, the changelog and the one line listing `radial_direction` in `docs/src/private_functions.md`.

Closes #115 · task `SymbolicAWEModels.jl-115`


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
   segments green-to-red by their spring force, and takes the colouring off again.
   `replay(log, sys; force_color=true)` starts with it ticked.
```
1-Bort-1, line 4: MINOR: The last sentence is a countermand — it only means something to a reader who knew the old warning, and the changelog's job is to state what is true now. Dropping it costs nothing; the first two sentences already say what changed.

1-Bort-1 (commented):

**Good**

- Scope matches the card exactly: one term in `spherical_spin`, the warning deleted, one extracted helper, one test — `git log`/diff show nothing else touched.
- Sign is right by construction: `apply_heading!` (transforms.jl:~165) rotates `heading - current` about `normalize(rel_pos)`, so `+turn_rate * radial_direction` is `+turn_rate` of heading rate.
- Implementation matches the field's existing docstring `"Angular velocity around radial axis [rad/s]"` (types.jl:1346) — no API or semantic drift.
- `radial_direction` extraction is genuinely behaviour-preserving: `rotate_around_y`/`rotate_around_z` (SymbolicAWEModels.jl:245) return a unit vector from a unit input, and the dropped `normalize` fed a `clamp`ed `dot` in `min_rotation`.
- No duplicate closed form left behind: `rotate_around_y` now has exactly one call site in `src/`, inside the new helper.
- Test is a finite difference against an independently-placed heading, not a restatement of the formula, and runs for both dynamics types — a wrong axis or magnitude fails it.
- Known limitation (PARTICLE wing `ω_b` still zeroed, pre-existing and equally true of `elevation_vel`) is stated in the card rather than silently worked around.
- New docstring is reachable from `docs/src/private_functions.md`, so Documenter `checkdocs` will not fail on it.
- `reposition!`'s docstring was updated alongside, so both entry points describe the same three-term spin.

**Not good**

- `CHANGELOG.md:9` — The last sentence is a countermand — it only means something to a reader who knew the old warning, and the changelog's job is to state what is true now. Dropping it costs nothing; the first two sentences already say what changed.
- Changelog entry sits under `### Added` though the user-visible change is behavioural (warn-and-rest → spin); `### Changed` fits better.
- A RIGID body's `ω_b` now carries the radial term too, but nothing asserts it — the FD test only pins `pos_w`/`vel_w`.
- For a `rot_point` transform whose heading reference body is not the rot object, the heading axis (`apply_heading!`) and the spin axis (`radial_direction`) differ slightly, so "turn_rate is the heading rate" holds exactly only in the wing case; pre-existing, unremarked.
- Test's `println` block is noise, but it matches every neighbouring testset in the file, so leaving it is the consistent choice.

1-Bort-1:
**Local full suite: PASS** (37 min, Julia 1.12.7, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `b86140f32`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34960231381/job/104351806936

1-Bort-1:
**Local full suite: PASS** (40 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1:
Fix the conflicts and failing CI

1-Bort-1:
Conflict fixed in b6b96240 (merge of main, CHANGELOG only). Windows is green since #348. The one red job, ubuntu 1.12 monolith, is #316's twist-alignment flake: same line, same static twist as its other failures, and no turn_rate in that model. Logged it on #316 and didn't touch the ramp here. CI is re-running on b6b96240.

1-Bart-1 (approved):

PR #343, opened 2026-09-15
system.yaml names a model's geometry and VSM settings, and the code reads them from there

### LINKED ISSUE #111: Updating/restructuring logic/number of input .yaml files
Streamline input by:

**bundling all settings in `settings.yaml`**
- [ ] integrate vsm_settings into `settings.yaml` and adjust relevant src code
- [ ] test workings

**Separate aerodynamic geometry definition**
- [x] rename files to `aero_geometry.yaml` and remove bridle lines from there 

**Separate structural geometry definition**
- [x] create file `struc_geometry.yaml` and add geometry information here
- [x] once added, continue in: https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/issues/59 using

1-Bort-1:
### TL;DR

`data/<model>/system.yaml` now names a model's `structural_geometry:`, `aero_geometry:` and `vsm_settings:` beside the `sim_settings:` it already had, and `project_file(entry)` resolves one of those against the data path — so src, the examples and the docs ask the project for a model's files instead of spelling the filenames out. It is the layout `V3Kite.jl`'s `data/system_psm.yaml` already uses, which is what the thread asked for.

### What was wrong

A model's filenames were written down in as many places as read them. `vsm_settings.yaml` was hard-coded in `SystemStructure`, in the precompile workload, in `docs/generate_figures.jl` and in seven examples; `data/2plate_kite/vsm_settings.yaml` carried `geometry_file: data/2plate_kite/aero_geometry.yaml`, a repo-relative path that only worked because `create_vsm_wing` re-anchored it with `basename()`; and `settings.yaml` carried `kite.struc_geometry_path` and `kite.aero_geometry_path`, which nothing in this package or in KiteUtils 0.12.2 ever read — `StructTypes.constructfrom!` drops unknown keys silently, so they were an earlier, dead attempt at this same issue. A second variant of a model therefore meant an edit everywhere a name appeared.

### What changed

`project_file(entry)` (`src/yaml_loader.jl`, exported and documented on the Parameters page) reads `system.<entry>` out of the project file and joins it to the data path, returning `""` where the project file does not exist or names no such entry. Every shipped `data/*/system.yaml` gained the entries its model actually has — `beam` and `kps4` still name only `sim_settings:`, and that keeps working.

`create_vsm_wing` now takes the project's `aero_geometry:` as a wing's geometry file where the wing names none of its own, so the name lives in one place, and `data/2plate_kite/vsm_settings.yaml` loses its `geometry_file:` line. A `VSMSettings` built outside a project still flies its own `geometry_file:`, absolute or relative as before. The auto-load in `SystemStructure` reads the project's `vsm_settings:` the same way. The two dead `settings.yaml` keys are gone from the shipped model, from the documented schema and from the eleven test fixtures that copied them.

`examples/sam_tutorial.jl` is the one caller that builds `VortexStepMethod.Wing(vsm_set)` itself rather than through `create_vsm_wing`, so it sets `vsm_set.wings[1].geometry_file` from the project explicitly. The two rigid-variant examples keep naming `rigid_structural_geometry.yaml` — that is a deliberate variant choice, not the model's default.

### Where I would push back

The issue's first checkbox asks for `vsm_settings` to be folded into `settings.yaml`. I did not do that, and I think it is the wrong move: `VSMSettings` is VortexStepMethod's schema, not KiteUtils', and `V3Kite.jl` resolved the same issue by pointing at it from `system.yaml` instead. Say the word and it becomes a second PR.

Two orphans in `data/2plate_kite` — `aero_geometry_old.yaml` and `aero_geometry_unconverged.yaml` — are referenced by nothing. Deleting them is a different concern from naming files, so they are still there.

Per-variant project files (V3Kite's `system_beam.yaml` shape) would let the two rigid examples stop naming their geometry too. Only two callers want it; not this diff.

### Verification

- [x] Reproduced first: n/a — this is a restructuring, not a bug
- [x] `test/test_project_file.jl` 9/9 pass (new; `project_file` did not exist before it)
- [x] `test/test_match_aero_sections.jl` 144/144 pass — red first: with `data/2plate_kite/system.yaml` back at its one-line `sim_settings:` form all three testsets error with `ArgumentError: WingSettings must specify either geometry_file or both obj_file and dat_file`, which is the aero geometry arriving from the project file and nowhere else
- [x] After merging `main` at ec826563 (d83c7313), which brought #308's chord-frame work and #355's span-order check: `test_project_file.jl` 9/9, `test_match_aero_sections.jl` 147/147 (now including `main`'s "a wing whose sections run from -y to +y is refused", whose fixture takes its aero geometry through the project file) and `test_live_polar.jl` 158/158. The only conflict was the two sides' entries under `## Unreleased` in `CHANGELOG.md`; both kept. The earlier merge of 23a22289 was checked the same way with `test_tether_init.jl` 58/58 and `test_tether_winch.jl` 99/99
- [x] `test/test_per_point_wind.jl` 33/33 pass — a fixture this diff edited, and a project file naming only `sim_settings:`
- [x] Examples: 14 of 17 run on e4159eb7, including all 7 this diff edits; the 3 that fail are in files it does not touch (#360)
- [ ] Local full suite on d83c7313: running (`20260918T125502Z`, Julia 1.12.7); `ce9d3bbc` passed (58 min). The run before it, on Julia 1.13.0, stopped at `test_helpers.jl:45` `!isfile("Manifest.toml")`: the worktree had no `Manifest-v1.13.toml`, so Pkg wrote a gitignored root `Manifest.toml`. `./bin/install -y +1.13` cleared it and `test_helpers.jl` then passed 9/9 on 1.13. Nothing on the branch changed
- [ ] GitHub CI on d83c7313: macOS 1.12, Documentation and reuse-lint pass; the rest running
- [x] Docs build clean locally before the merge (`include("docs/make.jl")`, 0 errors; only the pre-existing size-threshold and `@ref` warnings) and in CI on ce9d3bbc · up to date with `main` at ec826563
- Found on the way: the local suite can stop before it starts on `test_helpers.jl:50`'s mtime comparison of the `.default` manifests against `Project.toml`, which an install that stamps `Project.toml` later trips with no content change. That is #305, not this diff's idea to fix
- Benchmark: n/a
- Risk: a downstream project file that names no `aero_geometry:` while its `vsm_settings.yaml` also names no `geometry_file:` now fails at wing construction instead of at path resolution. Every model in this repo names one or the other.

### Scope

+154 / -71 across 37 files, of which 22 are one-or-two-line deletions of the dead `struc_geometry_path` / `aero_geometry_path` keys and the seven examples' switch to `project_file`. The logic is `src/yaml_loader.jl` (+17), `src/system_structure/wing.jl` and `src/system_structure/system_structure_core.jl`. Not stacked: #320 touches adjacent lines in `CHANGELOG.md`, `src/SymbolicAWEModels.jl` and `test/runtests.jl` but is a different concern.

Closes #111 · task `SymbolicAWEModels.jl-111`


src/yaml_loader.jl
```diff
@@ -8,6 +8,23 @@ using VortexStepMethod, KiteUtils
...
+
+Path of the file the project file names under `system.<entry>`, resolved against
+the data path. `entry` is `"sim_settings"`, `"vsm_settings"`, `"aero_geometry"`
+or `"structural_geometry"`. Returns `""` where the project file does not exist
+or names no such file.
+"""
+function project_file(entry, project=KiteUtils.PROJECT)
+    data_path = get_data_path()
+    project_path = joinpath(data_path, project)
+    isfile(project_path) || return ""
```
1-Bort-1, line 22: MINOR: Only one of the nine call sites (`system_structure_core.jl:927`) checks the `""` return; every other caller forwards it straight into `VSMSettings`/`load_sys_struct_from_yaml`. A project missing an entry, or a typo'd entry name, surfaces as `SystemError: opening file ""` instead of naming the file that is missing.

docs/src/tutorial_yaml.md
```diff
@@ -404,10 +404,9 @@ using SymbolicAWEModels, VortexStepMethod
 set_data_path("data/2plate_kite")
 set = Settings("system.yaml")
 vsm_set = VortexStepMethod.VSMSettings(
-    joinpath(get_data_path(), "vsm_settings.yaml"); data_prefix=false)
+    project_file("vsm_settings"); data_prefix=false)
 
-struc_yaml = joinpath(get_data_path(),
-    "rigid_structural_geometry.yaml")
+struc_yaml = project_file("structural_geometry")
```
1-Bort-1, line 409: MINOR: This swaps the tutorial's model from `rigid_structural_geometry.yaml` to the project's particle geometry, but the figure immediately below is still generated from the rigid one (`docs/generate_figures.jl:32`), and the card's stated policy is that rigid-variant callers keep spelling the filename out.

examples/sam_tutorial.jl
```diff
@@ -104,8 +104,9 @@ end
 # --- STEP 4: Add a kite ---
 
 vsm_set = VortexStepMethod.VSMSettings(
-    joinpath(get_data_path(), "vsm_settings.yaml");
+    project_file("vsm_settings");
     data_prefix=false)
+vsm_set.wings[1].geometry_file = project_file("aero_geometry")
```
1-Bort-1, line 109: MINOR: This hand-patches `wings[1].geometry_file`, re-implementing (for the first wing only) the resolution `create_vsm_wing` now does in a loop. Any user building a `Wing` directly from a project's `VSMSettings` has to know to repeat this line.

1-Bort-1 (commented):

**Verdict: APPROVE WITH COMMENTS** · 3 inline, 0 off the diff

**Good**

- `rg struc_geometry_path|aero_geometry_path` over the tree returns only the CHANGELOG entry recording their removal — the dead keys are gone from settings.yaml, all 11 fixtures and the docs schema.
- New `test_project_file.jl` is picked up by `runtests.jl`'s readdir filter, is registered in `KERNEL_SKIP_FILES` (so the skip-list validation passes), and restores the previous data path at the end.
- Card's claim about the lost vsm auto-load checks out: the three fixtures that copy `data/2plate_kite` then overwrite `system.yaml` with `sim_settings` only (test_flap_beam, test_timoshenko_joint, test_segment_nonlinear) build Body/joint models with no VSM wing.
- `data/base/vsm_settings.yaml` names no `geometry_file` and `data/base/system.yaml` names no `aero_geometry`, so the new empty-geometry fallback leaves that model byte-identical in behaviour.
- `examples/sam_tutorial.jl` is the only `VortexStepMethod.Wing(vsm_set)` caller outside `create_vsm_wing`, and it was patched when `geometry_file:` left `vsm_settings.yaml` — no silent breakage left behind.
- New public symbol landed complete: export, docstring, `@docs project_file` in `docs/src/parameters.md` (the only `@docs` for it), and a CHANGELOG entry.

**Not good**

- `src/yaml_loader.jl:22` — Only one of the nine call sites (`system_structure_core.jl:927`) checks the `""` return; every other caller forwards it straight into `VSMSettings`/`load_sys_struct_from_yaml`. A project missing an entry, or a typo'd entry name, surfaces as `SystemError: opening file ""` instead of naming the file that is missing.
- `docs/src/tutorial_yaml.md:409` — This swaps the tutorial's model from `rigid_structural_geometry.yaml` to the project's particle geometry, but the figure immediately below is still generated from the rigid one (`docs/generate_figures.jl:32`), and the card's stated policy is that rigid-variant callers keep spelling the filename out.
- `examples/sam_tutorial.jl:109` — This hand-patches `wings[1].geometry_file`, re-implementing (for the first wing only) the resolution `create_vsm_wing` now does in a loop. Any user building a `Wing` directly from a project's `VSMSettings` has to know to repeat this line.
- Test suite and docs build not run locally; this changes the path resolution every model load goes through, so CI green is the only evidence — do not merge red.
- The PR knowingly implements a different reading of #111 than the issue body's unchecked box asks; disclosed in the card, but it is Bart's call to confirm.
- `project_file` duplicates KiteUtils' `vsm_settings_file`/`aero_geometry_file`/`structural_geometry_file` with a different contract (resolved path, `""` guard); justified in the card, but two ways to ask the same question now exist.
- `project_file` re-parses the project YAML on every call and reads the mutable global `KiteUtils.PROJECT` rather than anything carried by the `set` in hand.
- It also skips the leading-`data/` stripping `KiteUtils.se` does, so an explicit `project_file(entry, "data/base/system.yaml")` silently returns `""` where `Settings("data/base/system.yaml")` works.
- The `basename()` flattening of a relative `geometry_file` survives untouched, so a wing naming `sub/foo.yaml` is still silently re-anchored to `<data>/foo.yaml`.
- A project file with `system:` present but empty makes `haskey(nothing, entry)` throw instead of returning `""`.
- The identical one-line header comment added to five `system.yaml` files says little the filename and `system:` key do not.
- `src/precompile_workload.jl:32` now takes the vsm path from global state while line 34 still takes the geometry from the `fixture` argument — correct only because `run_workload` sets the data path first.

1-Bort-1:
**Local full suite: PASS** (30 min, Julia 1.13.0, one cell of the matrix)

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `d313c3003`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34963849028/job/104363543852

1-Bart-1:
Did you run the examples to make sure they still work?

1-Bort-1:
Not before you asked. I have now: 14 of the 17 examples run on e4159eb7, including all 7 that #343 edits (sam_tutorial, coupled_linearize, coupled_2plate_kite, coupled_2plate_kite_linear_vsm, vsm_linearization, cold_start_inference, backend_scaling) and saddle_form, which reads the changed data/saddle_form/system.yaml. The other three fail in files #343 does not touch: heading_gate (ambiguous `save`), inflated_beam_fit (`ElasticJoint` keywords) and kps4_comparison (KiteModels is not in the examples env). I opened #360 for them.

1-Bart-1:
Address the review comments

src/yaml_loader.jl
1-Bort-1: Fixed in 3b7acccc: project_file now throws an ArgumentError naming the project file and the entry; the two callers where a missing entry is normal use the private optional_project_file, which returns nothing (an empty system: block included).

docs/src/tutorial_yaml.md
1-Bort-1: Fixed in 3b7acccc: the tutorial names rigid_structural_geometry.yaml again, matching the figure below it.

examples/sam_tutorial.jl
1-Bort-1: Fixed in 3b7acccc: sam_tutorial.jl builds its wing with create_vsm_wing(set, vsm_set), so the geometry resolution lives in one place.

1-Bort-1:
Done in 3b7acccc: the three inline comments and the precompile-workload point from the review body. The body points I left, and why, are under 'Review points left as they are' in the description of #343.

Issue #345, opened 2026-09-15
Windows CI is red: test_segment deletes a temp dir that load_log...


1-Bort-1:
Windows CI is red: test_segment deletes a temp dir that load_log still has mmapped

`Julia 1.12 - windows-latest - x64 - monolith` is red on `main` and has been since 01373cab (#339); it is the only red job, and the suite around it is healthy — 3536 pass, 1 errors.

The error is at test/test_segment.jl:724, in the testset's own cleanup: `IOError: rm("C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\jl_7OJTsj"): directory not empty (ENOTEMPTY)`.

The cause is the SysState round-trip block #339 added at test/test_segment.jl:700-721. `load_log` ends in `Arrow.Table(fullname)`, which reads the file through `Mmap.mmap`, and the returned `lg` is still alive when `rm(tmpdir; recursive=true)` runs. Windows does not unlink a mapped file, so the child delete is deferred and the parent rmdir finds the directory non-empty. Nothing about it is flaky: it fails every run on that platform and passes everywhere else, because only Windows holds the lock.

The repo already knows this and handles it in two places, so the fix is to adopt one rather than invent a third: test/util.jl:179-181 leaves the directory to Julia's exit-time cleanup and says why in the docstring, and test/test_beam_replay.jl:86-91 makes the `rm` best-effort in a `try`/`catch` on `Base.IOError`. My preference is util.jl's — a teardown that cannot succeed is better dropped than caught, and Julia removes the directory at exit anyway.

Worth fixing the class while there: any test that `load_log`s out of an `mktempdir` and then removes it has the same defect. The same run ends on a second one, `Failed to clean up temporary path "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\jl_eqle8F"`, which only warns because it comes from Julia's own atexit hook rather than a test.

Found from #338, whose Windows check is red for this and nothing of its own: the same branch was green on Windows at fa67ab88 and went red only when it merged main in at a1482050.

Job logs: main https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34957804412/job/104344176268 and #338 https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34958352560/job/104345728387

1-Bort-1:
[...]

PR #346, opened 2026-09-15
Stop the structural YAML loader on a field it does not read

### LINKED ISSUE #188: Throw on wrong field in yaml

1-Bort-1:
### TL;DR

A misspelled column in a structural YAML was dropped in silence and the component kept its constructor default, so a typo changed the model without saying so. The loader now carries a written-down schema of what each block accepts and errors on anything outside it, naming the block, the row and the fields that would have been legal.

### What was wrong

Three places dropped input on the floor. `call_yaml_constructor` iterates over the *constructor spec* and asks each named field of the row, so a column the spec does not name is never looked at — `extra_masss` left the point on `extra_mass = 0.0`. A top-level block nobody reads was never looked at either; only the three removed blocks (`materials`, `elements`, `segment_properties`) were rejected by name. And `parse_table` met a row with more values than the table has headers with a `@warn` and `continue`, so the component simply did not exist, and the failure surfaced much later as something else — in the fixture I wrote, as `No points or bodies found`.

### The schema had never been written down

The accepted set cannot be derived from the constructors. YAML column names are their own vocabulary (`wing_idx` becomes the constructor's `wing`, `point_idxs` becomes `points`), and about twenty fields — `sum_len`, `twist`, `set_value`, everything `load_yaml_bodies`, `load_yaml_joints` and the plate-wing loader read — never reach `call_yaml_constructor` at all. So `YAML_BLOCK_FIELDS` states it: block name to accepted field names, twelve blocks, checked once by `check_yaml_fields` right after the removed-block check. Retired names stay in the table so their own pointed error messages still win over the generic one. `load_yaml_joints` lost its `optional` argument and reads that set from the table instead, so the joint columns are now written once rather than twice.

### What wiring it up turned up

A point's `anchor_b` was in the table because `Point` takes it and `tutorial_yaml.md` documented it — and the loader never passed the column on, so a `BODY_STATIC` point took its body-frame offset from `pos_cad` whatever its row said. Fixed here, and `test_yaml_bodies.jl` now asks for a non-zero offset rather than the zero that passed either way. One limit stays: the resolver treats an all-zero `anchor_b` as unset (`iszero(point.anchor_b)` at `src/system_structure/system_structure_core.jl:1085`), so a row that deliberately writes `[0, 0, 0]` still gets the `pos_cad`-derived offset. The tutorial now says so. Making zero mean zero needs `Point` to carry a distinct unset state, which is a change to the core type rather than to the loader, so I have left it for a thread of its own if you want it.

Going the other way, the `idx` column the tutorial showed on points, segments, pulleys, tethers, winches and transforms is read by nothing (indices come from row order), and neither are a joint's `damping_trans` and `damping_rot`, which the constructors spell `damping`. The tutorial and the `test_yaml_weighted_ref.jl` fixture now call that column `name`, which every block accepts and which can be an integer, so they keep their numbering. The joint columns become `damping`. Any other YAML with an `idx` column now gets an error that names it and says ``Rename `idx` to `name`.`` — it is the one column every file written from the old tutorial carries — and the changelog entry says the same. The entry sits under Fixed, not as a breaking change: the column never did anything, so the only thing that changes is that a hidden no-op now surfaces.

`get_field_or_nothing` and `convert_to_type` were dead; they are deleted with their doc entries, since the diff had the file open anyway. (`parse_segment_type` was a third, which #338 has since removed on `main`.)

One find left alone: `docs/src/exported_types.md` lines 98 and 100 still list `damping_trans, damping_rot` on the joint constructors, which take `damping`. That is the Julia API, not the YAML one, and outside this diff's idea.

### Red checks along the way

The last `agent ci-local` red, `UndefVarError: set_unstretched_length! not defined` in `test/test_tether_init.jl:618`, came from a tree that changed under the run: it compiled the package from `2f9d8354` at 09:18Z, and my merge of `main` (which brought #349's `set_unstretched_length!` and its test) landed at 10:10Z, before the runner reached that file. On the merged tree the whole file is `Tether init_stretched_length Tests 58/58`. The earlier Windows red was #345, fixed on `main` by #348 and merged here, and the earlier `test_helpers.jl` red was #305's mtime comparison.

### Verification

- [x] Reproduced first: `test/test_yaml_fields.jl` against unchanged `src/yaml_loader.jl` is 1 pass / 3 fail — `No exception thrown` for the typo field and for the unknown block, and the long row gives `┌ Warning: Skipping row 1: has 4 values, expected 3` followed by an unrelated `No points or bodies found`
- [x] The `idx` hint red before, green after: `YAML fields` 4 pass / 1 fail (``Message: "Unknown field `idx` in row 1 of block `points` …"``, with no mention of `name`), then 5/5 (juliaserver, Julia 1.13.0)
- [x] On the tree merged with `main` at `c49a97c8`: `test_yaml_fields.jl` 5/5 · `test_yaml_weighted_ref.jl` 16/16 · `test_yaml_bodies.jl` 17/17 · `test_yaml_variables.jl` 18/18 (juliaserver, Julia 1.13.0). `test_tether_init.jl` 58/58 on the previous merge, `c1b6c97d`
- [x] Every structural YAML in `test/`, `examples/`, `data/` and `scripts/` passes `check_yaml_fields`, and so does every `yaml` block in `tutorial_yaml.md` except the multi-variable dict-row snippet, which borrows `dyneema` from the `variables:` block in the snippet above it. The two `vsm_settings.yaml` are VortexStepMethod's own format and never reach this loader
- [x] Up to date with `main` (merged at `c49a97c8`, bringing #308 and #355). Only `CHANGELOG.md` conflicted: both sides added Fixed entries under `## Unreleased`, and all four are kept. Neither PR touches `src/yaml_loader.jl` or the YAML tests
- [ ] Local full suite on `c49a97c8`: running (`agent ci-local`) — it posts its own result
- [ ] GitHub CI on `c49a97c8`: queued
- [ ] Docs build not run locally; `docs/src/private_functions.md` auto-merged (main's entries added, this branch's deletions kept) and CI's `Documentation` job builds it
- Benchmark: n/a — `check_yaml_fields` runs once per file load, off any stepping path
- Risk: a field that some YAML in the wild uses and that I read as unused. The table was built by reading every `yaml_field` call site rather than by sampling files, but a block-level loader I missed would now reject a file that used to load.

### Scope

+195 / -68 across 8 files against `main`. The table and its check are about 70 of the added lines and the new test 79; most of the rest is `docs/src/tutorial_yaml.md` and the `test_yaml_weighted_ref.jl` fixture renaming `idx` to `name`, against the two dead helpers coming out.

Not stacked: #343, opened after this branch was pushed, also edits `src/yaml_loader.jl` (adding `project_file` directly above the `get_field_or_nothing` this branch deletes) and `docs/src/tutorial_yaml.md`. The overlap is textual; whichever merges second rebases over a hunk boundary.

Closes #188 · task `SymbolicAWEModels.jl-188`


src/yaml_loader.jl
```diff
@@ -9,38 +9,74 @@ using VortexStepMethod, KiteUtils
...
+    for block in sort(collect(keys(data)); by=string)
+        haskey(YAML_BLOCK_FIELDS, block) ||
+            error("Unknown block `$block` in $yaml_path. Blocks: $blocks.")
+        table = data[block]
+        table isa AbstractDict || error("Block `$block` in $yaml_path must be " *
+            "a table with a `data` list, got $(typeof(table)).")
+        yaml_block_empty(data, block) && continue
+        accepted = YAML_BLOCK_FIELDS[block]
+        for (i, row) in enumerate(parse_table(table)), field in keys(row)
+            field in accepted || error("Unknown field `$field` in row " *
```
1-Bort-1, line 75: MINOR: `idx` was documented on six blocks in `tutorial_yaml.md` until this PR, so it is the field most existing user files carry, yet it gets only the generic "Unknown field" error while genuinely retired names (`type`, `init_unstretched_length`) are kept in the table for their pointed messages. Users hitting the one predictable migration get the least guidance.

docs/src/tutorial_yaml.md
```diff
@@ -230,25 +231,26 @@ points:
 | `fix_sphere` | Bool | false | Constrain the point to a sphere |
 | `fix_static` | Bool | false | Dynamically freeze the point position |
 
-A `BODY_STATIC` point rides a rigid body (`body:`, or `wing:` for a wing body)
-or a beam element (`joint:`); its body-frame offset is derived from `pos_cad`.
+A `BODY_STATIC` point rides a rigid body (`body:`, or `wing_idx:` for a wing)
+or a beam element (`joint:`). Its body-frame offset is `anchor_b` where the row
```
1-Bort-1, line 235: MINOR: "Its body-frame offset is `anchor_b` where the row gives one" is not true for an explicit `[0, 0, 0]`: the resolver only honours `anchor_b` when `!iszero` (system_structure_core.jl:1084), so a deliberate zero offset is still silently replaced by the `pos_cad`-derived one — the same silent-drop class this PR removes elsewhere.

1-Bort-1 (commented):

**Good**

- The reported bug is real and the fix is at the one place it belongs: `check_yaml_fields` runs once in `load_sys_struct_from_yaml` right after the removed-block check (src/yaml_loader.jl:742).
- I cross-checked `YAML_BLOCK_FIELDS` against every row access in the loaders (`call_yaml_constructor` specs + mappings, `load_wing`, `load_plate_wing`, `load_yaml_bodies`, `load_yaml_joints`, `parse_tether_init`, station/pulley/winch/tether post-sets) — no field read by a loader is missing from the table.
- No repo YAML regresses: both `data/2plate_kite/*_structural_geometry.yaml` and every inline `headers:`/dict-row block in test/ and docs/ use only accepted fields; `data/saddle_form/*` and `aero_geometry*.yaml` go through other readers.
- `anchor_b` now reaches `Point` (kwargs_spec at src/yaml_loader.jl:814) and the derive-from-`pos_cad` path is gated on `iszero(point.anchor_b)` (system_structure_core.jl:1084), so the changed `test_yaml_bodies.jl` assertion genuinely fails before the fix.
- `load_yaml_joints` losing its `optional` argument is a real de-duplication: the joint column list is now stated once, and the derived set matches the old explicit tuples exactly.
- The three deleted helpers have no remaining reference anywhere in src/, test/, examples/ or docs/, and `SegmentType` is kept on purpose as a deprecation, so nothing dangles.
- New test is picked up automatically by `runtests.jl`'s readdir discovery, carries the SPDX header, names what it protects, and the kernel-skip entry is required by the `KERNEL_SKIP_FILES` validation.
- Scope matches the card: `git diff --stat` is exactly +219/-101 across 8 files, with no change outside the stated idea.

**Not good**

- `src/yaml_loader.jl:75` — `idx` was documented on six blocks in `tutorial_yaml.md` until this PR, so it is the field most existing user files carry, yet it gets only the generic "Unknown field" error while genuinely retired names (`type`, `init_unstretched_length`) are kept in the table for their pointed messages. Users hitting the one predictable migration get the least guidance.
- `docs/src/tutorial_yaml.md:235` — "Its body-frame offset is `anchor_b` where the row gives one" is not true for an explicit `[0, 0, 0]`: the resolver only honours `anchor_b` when `!iszero` (system_structure_core.jl:1084), so a deliberate zero offset is still silently replaced by the `pos_cad`-derived one — the same silent-drop class this PR removes elsewhere.
- `YAML_BLOCK_FIELDS` is a hand-kept mirror of what the loaders read; nothing keeps them in sync, so a future column added to a loader is rejected until someone remembers the table.
- The long-row error from `parse_table` names neither the block nor the file, unlike the two errors beside it that name both.
- `YAML_BLOCK_FIELDS`'s docstring skips the signature line the repo's other documented constants use (`KERNEL_SKIP_FILES`).
- The new test repeats the mktempdir + `cp(data)` + `set_data_path` preamble that `runtests.jl:12-18` and two sibling YAML tests already do; a `util.jl` helper would now pay off.
- `docs/src/exported_types.md:98,100` still advertise `damping_trans, damping_rot` on the joint constructors — the card declines it as out of idea, but the same staleness was fixed in `tutorial_yaml.md` in this diff.
- Card's local full-suite box is unchecked; my field-by-field sweep says nothing else should break, but that is reasoning, not a green run.
- A top-level block added by a downstream `load_wing` method for a custom aero mode is now rejected unless it mutates this private const.

1-Bort-1:
**Local full suite: FAIL** (29 min, Julia 1.13.0, one cell of the matrix)

```
  World-frame panel alignment                             |    6             6      0.0s
  World-frame unrefined section alignment                 |    6             6      0.0s
  Segment Tests                                           |   91            91     11.2s
  Nonlinear segment spring                                |   15            15      9.7s
  validate_station_modes                                  |    7             7      0.5s
  STATIC twist on rigid VSM wing                          |    4             4      6.0s
  Tether init_stretched_length Tests                      |   58            58      2.4s
  Winch Tests                                             |   99            99     27.1s
  Timoshenko joint element                                |   30            30      9.0s
  Transform Tests                                         |  357           357     16.7s
  Tube rigidity laws                                      |  138           138      0.3s
RNG of the outermost testset: Random.Xoshiro(0xc498feeadd79d8e2, 0x3a9b66c56c77a494, 0x5fd387f83f5a61aa, 0xbd109cf331694834, 0x7f0125bab985a85c)
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Types.jl:68
    @ Pkg.Operations ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Operations.jl:3148
  [3] test
    @ ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/Operations.jl:3026 [inlined]
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:586
  [5] kwcall(::@NamedTuple{io::IOContext{IO}}, ::typeof(Pkg.API.test), ctx::Pkg.Types.Context, pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:562
  [6] test(pkgs::Vector{PackageSpec}; io::IOContext{IO}, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:172
  [7] test(pkgs::Vector{PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:161
  [8] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::PackageMode, subdir::Nothing, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:189
  [9] test()
    @ Pkg.API ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Pkg/src/API.jl:178
 [10] top-level scope
    @ none:1
 [11] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [12] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:310
 [13] _start()
    @ Base ./client.jl:577
```

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `6cf30b070`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34969004946/job/104380357475

test/test_yaml_weighted_ref.jl
```diff
@@ -26,40 +26,40 @@ using KiteUtils
 # Layout mirrors particle_structural_geometry.yaml.
 const WEIGHTED_REF_YAML = """
 points:
-  headers: [idx, pos_cad, type, wing_idx, transform_idx,
+  headers: [pos_cad, type, wing_idx, transform_idx,
```
1-Bart-1: why not rename idx to name, name can be integer

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
 
+## Unreleased
+
+### Fixed
+- `anchor_b` on a point row reaches the point. A `BODY_STATIC` point anchored to
+  a body took its offset from the geometry even when its row gave one, because
+  the loader never passed the column on.
+
+### Changed
+- BREAKING: a structural YAML naming a block or a row field the loader does not
```
1-Bart-1: thats not breaking thats a fix of a hidden bug

1-Bart-1 (changes requested):

src/yaml_loader.jl
1-Bort-1: Fixed in c1b6c97d: an `idx` column now errors with ``Rename `idx` to `name`.`` ahead of the field list, tested in `test/test_yaml_fields.jl`.

docs/src/tutorial_yaml.md
1-Bort-1: Right: `iszero(point.anchor_b)` at `src/system_structure/system_structure_core.jl:1085` treats an explicit `[0, 0, 0]` as unset. c1b6c97d makes the tutorial say so. Making zero mean zero needs a distinct unset state on `Point`, which is a core change rather than a loader one, so it is on the card as a find and not fixed here.

test/test_yaml_weighted_ref.jl
1-Bort-1: Done in 2f9d8354. The fixture and every tutorial example now use a `name` column with integer values, and the changelog tells people with an `idx` column to rename it to `name`. Since c1b6c97d the load error itself says so too.

CHANGELOG.md
1-Bort-1: Agreed. Moved it under Fixed without the BREAKING: tag in 2f9d8354.

src/yaml_loader.jl
```diff
@@ -9,38 +9,75 @@ using VortexStepMethod, KiteUtils
...
+this table stops the load.
 """
-function get_field_or_nothing(::Type{T}, row::NamedTuple,
-                               field::Symbol) where T
-    if !haskey(row, field) || yaml_unset(row[field])
-        return nothing
-    end
-    return convert_to_type(T, row[field])
-end
+const YAML_BLOCK_FIELDS = Dict{String, Vector{Symbol}}(
```
1-Bart-1, line 16: Do I understand it correctly that this code is very fragile, it is hand-derived from the yaml loader and any mismatch with the yaml loader is a hidden bug?

1-Bort-1: Yes. The table is a second copy of what the loaders read, and those reads are spread over src/yaml_loader.jl and src/aero_modes/plate.jl (drag_corr, surfaces). A mismatch fails one of two ways: a field the loader reads but the table lacks makes valid YAML error, which is loud and the first test catches; a field the table lists but the loader no longer reads is silently ignored again, which is the bug this PR fixes, now hidden. Every entry has a read site today, but nothing keeps it that way. The fix without a copy: wrap each parsed row so every read records the field, and error after the load on any field nothing read. The schema is then the loader itself. It touches every read site (~15 `hasfield(typeof(row), …)` checks become `haskey`). I recommend doing it on this PR and dropping the table; say go and I will.

PR #344, opened 2026-09-15
Write simulation results to an output folder instead of the repository root

1-Bort-1:
### TL;DR

Every result the package writes now lands under the new `get_output_path()` — `output` in the working directory by default — instead of in the working directory or in `data`. `data` holds only what a run reads, and the output folder holds only what a run writes, so deleting it is always safe.

### What was wrong

Two separate habits put results where nobody wanted them. Anything with a bare filename landed in whatever directory Julia happened to start in, which for everyone here is the repository root: the replay viewer's Save button (#227), `record(lg, sys, "sim.mp4")`, the tutorial's three `.gif`s, `heading_gate.jl`'s PDFs and `errors.txt`, `logo.jl`'s PNGs and SVG, `reference_frames.jl`'s PDF. Anything going through `save_log` landed in the **data** folder, because KiteUtils' `save_log(logger, name)` falls back to `DATA_PATH[1]` when no `path=` is given — so `sim!`, `sim_reposition!` and every example dropped their `.arrow` next to the YAML the run had read. `.gitignore` carrying a line for `data/tmp.arrow` is the scar from that.

Uwe settled the shape on the thread: the output folder holds simulation results, it is safe to delete because they can be recreated, and `data` is configuration, i.e. input.

### What the branch does

`src/output_path.jl` is the whole mechanism — `OUTPUT_PATH`, `set_output_path`, `get_output_path` — mirroring KiteUtils' `DATA_PATH` idiom so the pair reads next to the `get_data_path`/`set_data_path` the package already re-exports. KiteUtils has no output-path concept of its own (I searched `OUTPUT`, `output_path`, `get_output` across its `src/`), so this is new rather than duplicated; it would arguably sit better *in* KiteUtils next to the data path, for KiteControllers and the viewers to share, but that would hold this change behind a KiteUtils release, so I have left it here and would move it on request.

`get_output_path()` creates the folder and returns it. A getter with a side effect is the one thing here worth arguing about, and it is deliberate: it deletes the `mkpath(get_data_path())` lines that `sim!` and `sim_reposition!` were carrying, and it means no caller can get the folder without it existing — which is what makes "delete the folder whenever you like" true rather than aspirational.

[...]

PR #348, opened 2026-09-15
Leave a replayed log's temp dir to Julia's exit-time cleanup

### LINKED ISSUE #345: Windows CI is red: test_segment deletes a temp dir that load_log...

1-Bort-1:
### TL;DR

`load_log` memory-maps its Arrow file, and Windows will not unlink a mapped file, so every test that loaded a log out of an `mktempdir` and then `rm`'d that directory was deleting a directory it still held open. The four teardowns that did this now leave the directory to `mktempdir`'s own exit-time cleanup, which is what `test/util.jl` already did.

### What was wrong

`Julia 1.12 - windows-latest - x64 - monolith` has been the only red job on `main` since 01373cab, with the suite otherwise healthy: the run ends `3536 pass, 1 errors`, and the one error is a teardown, not an assertion.

    IOError: rm("C:\Users\RUNNER~1\AppData\Local\Temp\jl_7OJTsj"): directory not empty (ENOTEMPTY)

### Why it happened

`load_log` (KiteUtils v0.13.0, `src/load_log.jl:38`) ends in `Arrow.Table(fullname)`, and `Arrow.Table` reads the file through `Mmap.mmap`. The columns of the returned `SysLog` are views onto that mapping, so the mapping stays alive exactly as long as the log does — which, in each of these testsets, is past the `rm`.

I measured the mapping rather than inferring it: after a `save_log`/`load_log` pair into a temp dir, the process holds one live mapping on the `.arrow` file while the returned log is still in scope.

On POSIX that is harmless — the unlink succeeds and the inode goes when the mapping does — which is why the suite is green everywhere else. Windows defers the child delete instead, so the parent `rmdir` that follows finds the directory non-empty. Nothing about it is flaky; it fails every run on that platform and only on that platform.

### What I changed

The defect is the class, not the line #339 added, so I fixed it as a class: any test that `load_log`s out of an `mktempdir` and then removes it has the same bug, whether or not it has fired yet.

    test/test_segment.jl        the red one — a bare rm
    test/test_beam_replay.jl    rm wrapped in try/catch on Base.IOError
    test/test_makie_extension.jl  a bare rm, GLMakie-only so never red on CI
    test/test_for_precompile.jl   rm in a finally; excluded from runtests.jl

Those four are the whole class, not a sample: `test/` has 26 `rm(tmpdir)` teardowns, but only five files call `load_log` at all, and the fifth is `util.jl`, which already left its directory alone. The other 21 write YAML fixtures and map nothing, so their `rm` is correct and stays.

All four now drop the `rm`. The repo already held two different answers to this — `test/util.jl` left the directory alone and explained itself, `test_beam_replay.jl` caught the error — and I took `util.jl`'s rather than adding a third: a teardown that cannot succeed is better dropped than caught, and `mktempdir` registers the directory for removal at exit anyway. That also means the `try`/`catch Base.IOError` in `test_beam_replay.jl` goes, which is the only place a real IO error was being swallowed.

One line at each site says why there is no teardown, because the next reader's instinct on seeing an `mktempdir` with no `rm` is to add one back. `util.jl` carried that reason as a docstring paragraph; it moves to the same one-liner at the call site, since a docstring says what and not why (CLEAN_CODE §4).

### Where I would push back

Julia's own atexit hook still warns `Failed to clean up temporary path` on Windows, for the same reason — the same CI run already ends on one of those. That is the end state this change accepts, not a regression it introduces: a warning from the runtime rather than an error from a test, and the directory is the OS's to reap. Making it silent would mean getting the mapping released before exit, which is `load_log`'s to offer and KiteUtils' to change.

Reviewing the same files turned up a sibling of this bug that I have not touched, because fixing it would change what a test exercises rather than only what it cleans up. `build_test_syslog` in `test_makie_extension.jl` always saves under the name `makie_test_log`, and it is called twice while the first log is still live: `lg1` maps `<tmpdir>/makie_test_log.arrow`, then the second call re-writes that same path, and `lg1` is read six more times afterwards. `save_log` goes to `Arrow.write` → `open(file_path, "w")`, so that is a truncate of a file this process has mapped — the same conflict as the `rm`, one step short of a delete. Windows refuses it; POSIX quietly mutates the bytes `lg1` is viewing, which nothing notices because the two logs are near-identical. It has never gone red because the file is gated on `GLMAKIE_AVAILABLE`. A distinct log name per call fixes it, but whether the assertions still mean the same thing once `lg1` and `lg2` are genuinely different files is a question for the reviewer, not something to fold in here silently.

Separately, `validate_sysstate_roundtrip` in `test/util.jl` still has four paragraphs of rationale in its docstring after I took the fifth out. I left them: they are pre-existing and nothing to do with this bug, and trimming them would make this diff harder to read for no gain. Worth a `cleanup:` pass with the rest of that file.

### Verification

- [x] Reproduced first: not locally — the `rm` succeeds on Linux. Reproduced in CI, where it is 100% on windows-latest and green on every other platform: the `main` run, and #338's check, which is red for this and nothing of its own.
- [x] `test/test_segment.jl` 91/91 · `test/test_beam_replay.jl` 7/7 · `test/test_makie_extension.jl` 21/21 (GLMakie live) · `test/test_for_precompile.jl` clean — juliaserver, exit 0
- [x] Local full suite (`agent ci-local`): PASS twice — 32 min and 40 min, Julia 1.12.7, exit 0. The second finished after the second commit landed, so it spans the final code; a third run on the exact final tree is in flight.
- [x] Up to date with `main` · no changelog entry: test-only, nothing user-visible ships · no new or renamed public symbol, so no docs page
- [ ] REUSE lint: **not run** — `bin/reuse_lint` reports the `reuse` tool is not installed on this box
- [x] **GitHub CI, windows-latest: green.** `CI.yml` triggers on `pull_request` only, so a `workflow_dispatch` run carried the job before this PR existed: `Julia 1.12 - windows-latest - x64 - monolith` passed in 1h31m42s, against red on every run of `main` since 01373cab. macOS and Documentation are green in the same run; the ubuntu cells and reuse-lint were still queued when this was written, and they are green on `main` already.
- That run is on `dc35cbd8`. The later `81335c2e` only moves two comments to the teardown sites they replace, so no executable line differs between them; this PR's own run covers the final commit.
- Risk: the fix is the absence of a teardown, so nothing can fail at the site any more; what it cannot prove is that no *later* test depends on one of these directories being gone. None reads a stale data path today, and the local full suite passing is the evidence for that.

### Scope

+5 / -15 across 5 test files; no `src/` change. Not stacked: #343 touches `test_segment.jl` and `test_makie_extension.jl` and #344 touches `test_beam_replay.jl` and `test_makie_extension.jl`, but both are themselves red on Windows for this bug, so this goes to `main` ahead of them rather than behind. #344 inserts a testset immediately above the `test_beam_replay.jl` teardown this removes, and both touch the same region of `test_makie_extension.jl` — whichever merges second resolves by keeping the insert and dropping the `rm`.

Closes #345 · task `SymbolicAWEModels.jl-345`


test/test_for_precompile.jl
```diff
@@ -48,7 +48,7 @@ syslog = try
     save_log(logger, "_hanging_mass")
     load_log("_hanging_mass")
 finally
-    rm(tmpdir; recursive=true)
+    # No teardown: load_log mmaps the Arrow file, and Windows locks a mapped file.
```
1-Bort-1, line 51: MINOR: The note reads "No teardown" while sitting directly above `set_data_path(prev_data_path)`, which is a teardown — the block still cleans up, just not the directory. Costs a reader a second pass to see which teardown is meant.

1-Bort-1 (commented):

**Verdict: APPROVE** · 1 inline, 0 off the diff

**Good**

- The class claim checks out: `grep -rln load_log test/` returns exactly the 5 files touched, and the other 21 `rm(tmpdir)` teardowns write only YAML/MP4 fixtures, so leaving them is right.
- Root cause is the real one: `load_log` → `Arrow.Table` → `Mmap.mmap`, and the returned columns keep the mapping alive past the `rm`, which is why only Windows goes red.
- Converging on `util.jl`'s answer instead of keeping the `try`/`catch Base.IOError` in `test_beam_replay.jl` removes the repo's second codepath and stops swallowing genuine IO errors (CLEAN_CODE §2).
- Moving the Windows rationale out of `validate_sysstate_roundtrip`'s docstring into a one-liner at the call site is the right side of §4 (docstrings say what, not why).
- Scope is one idea, test-only, +5/-15, no `src/` change, no changelog needed — confirmed against the diff.
- The known sibling bug (`build_test_syslog` reusing `makie_test_log` while `lg1` is still mapped) is written up in the card rather than silently folded in or ignored (§5).

**Not good**

- `test/test_for_precompile.jl:51` — The note reads "No teardown" while sitting directly above `set_data_path(prev_data_path)`, which is a teardown — the block still cleans up, just not the directory. Costs a reader a second pass to see which teardown is meant.
- The same sentence is now pasted verbatim at five sites; defensible per-file, but it is five copies of one fact that will drift if the reason ever changes.
- In `test/util.jl:188` the note sits after `load_log`, not at `path = mktempdir()` on line 183 where a reader's instinct to add cleanup actually fires — and here it replaces no teardown at all.
- `validate_sysstate_roundtrip` runs per `test_init!`, so a full suite now retains one temp dir per model until exit; harmless but the accumulation is larger than the four sites suggest.
- REUSE lint was not run (tool absent); no files were added or moved, so nothing here can plausibly trip it.
- The Makie sibling bug arguably meets the bar for its own issue rather than a card bullet, since fixing it changes what the assertions exercise.

1-Bort-1:
**Local full suite: PASS** (30 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

PR #349, opened 2026-09-15
Let a tether's unstretched length be set, and split reinit! into steps a caller can run

### LINKED ISSUE #347: A tether's rest length cannot be set: reinit! recomputes it from the node positions
A tether's rest length cannot be set. `tether.len` is documented as "Unstretched tether length [m] (sum of segment l0). ODE state variable" (`system_structure/types.jl:984-986`), but assigning it does not survive: `reinit!(sys_struct, set)` derives the segments from the node positions and then rewrites `len` from them (`system_structure/utilities.jl:753-773`), so whatever the caller wrote is replaced by the geometry's own length.

Measured in V3Kite.jl: `build_settling_struct` asks for a 247.557 m tether and for a 240.0 m tether, and both come back as exactly 250.0 m — the length in the geometry YAML. Writing `tether.len` after `load_sys_struct_from_yaml`, and again after the state restore that follows it, changes nothing: the first replay step still reads 249.99999999962404 m, bit-identical across runs.

The only way to place a tether at a wanted length today is `tether.init_stretched_len` plus `init!(...; reinit_sys=true)`, which runs `apply_tether_init_stretched_lens!` (`utilities.jl:562`). That works, but it is a whole-structure re-placement: it resets every body's pose to `pos_cad`, resets station twist, and recaptures each joint's rest configuration from the placed poses (`utilities.jl:734-748, 840-843`). A caller that only wants a different tether length has to copy its current pose into `pos_cad` first to keep it — which then redefines what "rest" and "CAD" mean for every beam joint and every station flap deflection, since `init_joint_rest!` and `init_station_flap!` read exactly those fields (`rigid_body.jl:589-598, 696-697`).

What this costs downstream: a run that settles a kite and then replays recorded tether lengths hands over a tether of one length and imposes another on the first step. In V3Kite.jl's beam replay that is a 2.44 m step on a 248 m rope, and the winch force goes 606 N -> 6704 N in one 50 ms step, then to 3.5 N — a slack rope — and rings for about two seconds before the run settles down.

What I would like: a way to set a tether's unstretched length that holds, without re-placing the structure — either `tether.len` being honoured as its docstring says (`reinit!` taking the assigned value instead of recomputing it when one is set), or an explicit `set_tether_length!(sys_struct, tether, len)` that rewrites the segment `l0` and leaves poses, joint rests and station flap references alone. Either would let a caller move the tether length per step, which is what a data replay does.

Happy to write it if you say which of the two shapes you want.

1-Bort-1:
### TL;DR

`set_unstretched_length!(sys_struct, tether, len)` sets a tether's rest length and moves nothing else, and `reinit!(sys_struct, set)` is now a sequence of nine exported steps. That gives two documented routes: `init!(sam; reinit_sys=true)` applies every adjustment, or run the steps you want on `sam.sys_struct` and then `init!(sam; reinit_sys=false)`. Before this, the only way to set a tether's length was `init_stretched_len` plus a full re-placement, which also resets every pose, joint rest and flap reference.

### What was wrong

`reinit!` derives `tether.len` from the placed geometry, so a length assigned by hand was overwritten. Reproduced on the Route-2 fixture from `test_tether_init.jl`: assign `tether.len = 150`, and the segments' `l0` stay at `[100, 100]`. After `reinit!`, `len` is back at 200. The only other way to get a chosen length re-places the whole structure.

### What changed

- `set_unstretched_length!` writes `tether.len` and shares it equally over the segments' `l0`. `apply_tether_init_forces!` now calls it, so `reinit!`'s own copy of that loop is gone.
- `reinit!(sys_struct, set)` now just calls these steps in order: `reset_to_cad!`, `apply_tether_init_stretched_lens!`, `update_segment_lengths!`, `apply_tether_init_forces!`, `init_pulley_lengths!`, the transforms' `reinit!`, `remake_wing_aero!`, `init_wind!`, `relax_segments!`, `init_rest_geometry!`. Its docstring lists the order and which kwarg skips which step. The steps are exported and listed under a new "Placing a SystemStructure" section of the Functions page, with the example of setting a tether length.
- Two copies of the same logic are gone, with no change in behaviour. `reinit!` reset each body pose by hand, then called `copy_cad_to_world!`, which reset it again. The `SystemStructure` constructor had its own copy of the `ignore_l0` loop, which is now `relax_segments!`. That kwarg's docstring said "from the CAD geometry" but the loop reads `pos_w`, so the docstring now says "placed length".

### Where I would push back

- `init!` keeps `ignore_l0`, `apply_tether_lengths`, `reset_vel` and `remake_vsm`, and `reinit!` keeps `apply_transforms`. Each of them now just runs or skips one step, so with step-by-step calls they are redundant. Removing them is BREAKING and touches `reset_vel=false` in about ten test files and `remake_vsm=false` in two examples. `remake_vsm` also defaults to `true` on `init!` but `false` on `reinit!`. I left them in; say if they should go, in a follow-up.
- `reset_to_cad!` also sets winch reel-out speeds back to `init_vel`, which is not strictly CAD. I kept it there rather than add a one-loop step of its own.
- Not touched: `wing_kinematics_from_points!` (`src/aero_modes/common.jl:1164-1166`) repeats `init_wind!`'s profile-wind line, but it does it per step and per aero point, so the two do not share a body cleanly.

### Verification

- [x] Reproduced first: `tether.len = 150` left `l0` at `[100, 100]`, and `reinit!` put `len` back to 200.
- [x] `test/test_tether_init.jl` "set_unstretched_length! sets len without re-placing": red before (`UndefVarError: set_unstretched_length!`), green 58/58 after.
- [x] `test/test_tether_winch.jl` "set_unstretched_length! retrims a winched tether" (route 2 on a winched tether, pose bit-identical): red before, green 99/99 after.
- [x] The `reinit!` split adds no test of its own because it changes no behaviour. Every file below that runs `reinit!` passes with the same counts as before the split (juliaserver): tether_init 58, tether_winch 99, pulley 82, transform 325, rigid_body 25, section_alignment 19, heading_calculation 152, per_point_wind 33, flap_beam 28, joint 37, wing_dynamics 44, static_twist 11, wing 171.
- [x] Docs build clean (no new warnings) · up to date with main (merged for #348)
- Local full suite: on the previous head (0cc401f1) PASS; on f2383d74 not finished when this was written · GitHub CI: pending on f2383d74
- Benchmark: n/a
- Risk: `set_unstretched_length!` errors on `len <= 0`, so a tether whose endpoints coincide now errors inside `reinit!`. That configuration already failed `validate_sys_struct` one step later (`segment.l0 <= 0`).

### Scope

+261 / −117 across 10 files. `utilities.jl` accounts for +236/−: the setter and the nine step functions, each with a docstring, replacing a 110-line `reinit!` body. The rest is exports, the docs section, the changelog and the two tests.

Closes #347 · task `SymbolicAWEModels.jl-347`


1-Bort-1 (commented):

**Good**

- Fold is behaviour-preserving: read `reinit!` around line 789 — the deleted loop ran unconditionally right after `apply_tether_init_forces!`, and the new call site sits in the same place with the same empty-tether skip, before the pulley `sum_len` block.
- One source for the len→l0 mapping now: `rg segment_idxs`/`sum_len` shows the only remaining eager struct write is in `set_unstretched_length!`; the other three splits are model equations or pre-`SystemStructure` sizing, as the card claims.
- `tethers` dropped from `reinit!`'s destructuring because nothing reads it any more — no leftover unused binding.
- Docstring countermand deleted (§7): `types.jl` no longer says `len` "does not survive reinit!" but states what is true and links the setter with `@ref`.
- New exported symbol lands with export, `docs/src/exported_functions.md` entry and a changelog bullet in a fresh `## Unreleased` section — `checkdocs = :all` will be satisfied.
- Tests name the behaviour and pin the point of the feature: positions and `pos_cad` unchanged, `l0` split, `reinit!` overwriting again, and the winched path reaching `u0` via `init!(reinit_sys=false)`; both error branches covered.

**Not good**

- `test/test_tether_init.jl:619` — This testset reads `r1_yaml.yaml` but never writes it — the file only exists because Test 2 wrote it earlier in the same file. Reordering, skipping or rewriting Test 2 turns this into a file-not-found failure that looks unrelated to the setter.
- `src/system_structure/utilities.jl:674` — The docstring of the exported setter never says how the new length reaches a running simulation — that a winch-driven tether's length is integrator state and needs `init!(sam; reinit_sys=false)`, while a winchless one is a parameter picked up at the next step. That is the one thing a caller needs, and it currently lives only in the PR card and a test comment.
- For a tether whose segments also belong to a pulley, the setter writes `l0` but leaves `pulley.sum_len` stale, and the model takes the pulley equation for those segments — so such a tether retrims only partially, unmentioned anywhere.
- `isempty(tether.segment_idxs)` error is unreachable from the only in-repo caller, which already `continue`s on it; it is untested and duplicates the guard.
- `unstretched_length(sam)` reads from a `SymbolicAWEModel` while the setter takes `(sys_struct, tether)`; the getter/setter pair is asymmetric for callers holding a `sam`.
- test/test_tether_winch.jl:651-653 leaves `winch.brake` and `seg.unit_stiffness/unit_damping` mutated on the shared fixture; safe only because it is the last testset.
- test/test_tether_winch.jl:660-661 is a two-line inline comment where the rubric asks for one line.
- The changelog bullet runs five lines and re-explains `reinit!` semantics that the docstring already carries.

test/test_tether_init.jl
```diff
@@ -612,5 +612,38 @@ winches:
             sys.segments[:tail_seg], sys.points) ≈ 1.0
     end
 
+    # ================================================================
+    # Test 12: set_unstretched_length! writes len and l0 and moves nothing
+    # ================================================================
+    @testset "set_unstretched_length! sets len without re-placing" begin
+        yaml_path = joinpath(tmpdir, "r1_yaml.yaml")
```
1-Bort-1, line 619: MINOR: This testset reads `r1_yaml.yaml` but never writes it — the file only exists because Test 2 wrote it earlier in the same file. Reordering, skipping or rewriting Test 2 turns this into a file-not-found failure that looks unrelated to the setter.

src/system_structure/utilities.jl
```diff
@@ -666,17 +666,38 @@ function init_unstretched_len(tether, segments)
     return stretched * (1 - force / stiffness)
 end
 
+"""
+    set_unstretched_length!(sys_struct::SystemStructure, tether::Tether, len)
+
+Set `tether`'s unstretched length [m] and share it equally over its segments'
+`l0`. Point positions, body poses, joint rest geometry and station flap
+references are left as they are.
```
1-Bort-1, line 674: MINOR: The docstring of the exported setter never says how the new length reaches a running simulation — that a winch-driven tether's length is integrator state and needs `init!(sam; reinit_sys=false)`, while a winchless one is a parameter picked up at the next step. That is the one thing a caller needs, and it currently lives only in the PR card and a test comment.

1-Bort-1:

1-Bort-1:
**CI: `Julia 1.12 - windows-latest - x64 - monolith - pull_request` failed** on `0cc401f17`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/34997778026/job/104478259815

1-Bart-1:
So now the API should become, and be documented as such:

1. use `init!` with `reinit_sys=true` to apply all adjustments
2. split up the `reinit!` function into runnable components like done for the tether length adjustment, and run the components wanted before running `init!` with `reinit_sys=false`

This avoids needing lots of kwargs on `init!`.

1-Bort-1:
Done in f2383d74 on #349: reinit!(sys_struct, set) is now nine exported steps, documented both ways under "Placing a SystemStructure". init!'s existing kwargs are kept; removing them is breaking, so it's a question on the PR.

1-Bart-1 (approved):

PR #351, opened 2026-09-16
Set the julia compat to 1.12 and 1.13, the versions CI tests

### LINKED ISSUE #350: Julia 1.13: SymbolicAWEModels.jl, compat to 1.12
OpenSourceAWE/SymbolicAWEModels.jl#322 moved the matrix to 1.12 and 1.13 and
tracks a `.default` for each, but `[compat]` on `main` still says
`julia = "1.11, 1.12, 1.13"`: it claims a Julia nothing tests. Make it
`"1.12, 1.13"`, as KiteUtils.jl has it.

A compat edit changes the project hash both manifests record, so re-resolve each
`.default` under its own Julia by the minimal resolve (`AGENTS.md` §2, and step 3
of the shared procedure for starting the session on 1.13), and commit them with
`Project.toml`. Check that nothing in `bin/`, the workflows or the README still
names 1.11; anything that does moves in this pull request. Nothing else changes.

Unit `symbolicawe-compat` of plan `plans/julia-1-13.md`, tracked in 1-Bart-1/Agents#282.
https://github.com/1-Bart-1/Agents/blob/bd9d5bf4f48b8dd00a025774b80762f9f2114c08/plans/julia-1-13.md

1-Bort-1:
### TL;DR

`[compat]` now says `julia = "1.12, 1.13"` instead of `"1.11, 1.12, 1.13"`, the same as KiteUtils.jl. CI stopped testing 1.11 in #322, so the old line claimed support for a Julia version nobody tests.

### What moved

`Project.toml`'s `julia` entry, one `Changed` line in the changelog, and both tracked manifests. Each `.default` went through the minimal resolve under its own Julia: 1.12.7 for `Manifest-v1.12.toml.default` and 1.13.0 for `Manifest-v1.13.toml.default`. In both, the resolve changed exactly one line: the manifest's record of SymbolicAWEModels' own version, from `0.16.0` to `0.18.0`. No dependency moved.

`project_hash` did not change in either manifest: it is `26716ea7…` before and after. Pkg leaves the `julia` compat entry out of that hash, so the issue's premise is wrong. Only the stale package version needed writing back.

Since then `main` released v0.18.1, and this branch merged it. The only conflict was the changelog: the Julia entry now sits under a new `Unreleased` section above v0.18.1. The release changed nothing in `Project.toml` except `version`, so the manifests still record `0.18.0` as the package's own version. They were one release behind on `main` too, and `bin/install` handles that without trouble.

A search for `1.11` in `bin/`, `.github/workflows/` and the README found nothing. The CI matrix and the `bin/install` menu already offer only 1.12 and 1.13, so none of those files changed.

### Found on the way

- `Artifacts.toml.default` still lists `models_v1_10` and `models_v1_11`, the model tarballs from the v0.5.0 release. The file it seeds, `Artifacts.toml`, is gitignored, and nothing in `src/`, `bin/` or the tests refers to it. The issue said nothing else changes, so it stays here. I would delete it in a `cleanup:` PR.
- Step 3 of the plan's shared procedure, `JULIAUP_CHANNEL=1.13 juliaserver launch .`, starts a **1.12.7** session. `juliaserver` starts Julia in a tmux shell, and that shell takes the tmux server's environment, not the caller's. What worked was running `exit()` in this worktree's own pane, then `julia +1.13 --project=<worktree> -t auto` there. The other `julia-1-13` units will hit the same thing, so the procedure in 1-Bart-1/Agents needs this fix.

### Verification

- [x] Reproduced first: n/a, compat metadata. Before the change, `Project.toml:103` read `julia = "1.11, 1.12, 1.13"`.
- [x] Minimal resolve under 1.12.7 and under 1.13.0 (juliaserver): "No packages added to or removed from" for both manifests.
- [x] `using SymbolicAWEModels` on 1.12.7 after the resolve: loads, v0.18.0.
- [x] Merged `origin/main` at `d51701fa` (v0.18.1); the search for `1.11` in `bin/`, the workflows and the README still finds nothing.
- [x] Local full suite on `94a62cba` (`agent ci-local`, Julia 1.12.7): PASS, `3574/3574`, 36 min, exit 0. The first run on the merge stopped at `test/test_helpers.jl:50`, `mtime(Manifest-v1.12.toml.default) >= project_mtime - 1.0` evaluating to `1.7895700061e9 >= 1.7895865057e9`: the merge rewrote `Project.toml` after the `.default`s were written. That check compares timestamps rather than contents (#305), and the tracked diff against `origin/main` is the four files above. I touched both `.default`s and re-ran.
- [x] GitHub CI on `94a62cba`: all seven checks pass — 1.12 monolith on ubuntu, Windows and macOS, 1.12 kernel, 1.13 monolith, docs, REUSE lint. The 1.13 suite was not run locally; that cell covers it.
- Benchmark: n/a
- Risk: none beyond a 1.11 user losing the install, which is the point.

### Scope

+9 / −3 across 4 files: `Project.toml` (1 line), one line in each `.default`, and a 6-line changelog entry.

Closes #350 · task `SymbolicAWEModels.jl-350`


1-Bort-1 (commented):

**Good**

- The PR does exactly what the card plans: one `julia` compat line, one changelog entry, and one line in each manifest. `git diff --stat origin/main...HEAD` shows +9/−3 across 4 files, matching the card's Scope.
- The manifest change from 0.16.0 to 0.18.0 is the only line that moved. It matches `Project.toml` `version = "0.18.0"`, and the card names it as a result of the resolve, so it is expected housekeeping, not scope creep.
- No stale `1.11` claim is left behind: `rg 1\.11` outside the manifests finds only earlier released changelog sections, CSV data and `Documenter = "1.11"` in `docs/Project.toml`.
- The `Unreleased` header is new and sits above `v0.18.0`, the release made the day before, so the entry is not stuck inside a released section.
- Both 'Found on the way' items are written down: the leftover `Artifacts.toml.default` entries and the juliaserver/tmux environment problem. §5 asks for exactly that, and neither is quietly worked around.

**Not good**

- `CHANGELOG.md:6` — This changelog raised the Julia requirement once before as `BREAKING: Julia version requirement raised from 1.10 to 1.11, 1.12.` Dropping 1.11 is the same kind of change, but this entry has no `BREAKING:` prefix, so it is easy to miss when Bart picks the next version number.
- In Pkg, `julia = "1.12, 1.13"` means `^1.12`, so it accepts every 1.x from 1.12 on. The changelog's 'installs on Julia 1.12 and 1.13' reads as tighter than it is, but it matches KiteUtils, so no change is needed.
- 'the two versions CI tests' in the changelog is rationale. That belongs in the PR, and deleting it leaves the entry just as clear (§3/§5).
- The `Artifacts.toml.default` leftover is only a sentence in the card. A tracked issue or a follow-up `cleanup:` PR would keep it from being forgotten after the merge.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unreleased
+
+### Changed
+- Julia 1.11 is no longer supported: the package installs on Julia 1.12 and 1.13,
```
1-Bort-1, line 6: MINOR: This changelog raised the Julia requirement once before as `BREAKING: Julia version requirement raised from 1.10 to 1.11, 1.12.` Dropping 1.11 is the same kind of change, but this entry has no `BREAKING:` prefix, so it is easy to miss when Bart picks the next version number.

1-Bort-1:
**CI: `Julia 1.12 - ubuntu-latest - x64 - kernel - pull_request` failed** on `9daff2ade`.

https://github.com/OpenSourceAWE/SymbolicAWEModels.jl/actions/runs/35115320808/job/104859142817

1-Bort-1:
**Local full suite: PASS** (26 min, Julia 1.12.7, one cell of the matrix)

1-Bart-1 (approved):

PR #355, opened 2026-09-16
Drop the per-panel span sign from the panel force path

### LINKED ISSUE #353: Drop the reversed-span-order handling from the panel force path:...
Drop the reversed-span-order handling from the panel force path: panel_span_signs, orient and the :_down kernel

Bart on #308 (comment 5700578177): assume the spanwise ordering VSM uses, +y to -y, error if it is not that order, and remove the code that tries to deal with the opposite ordering.

#308 adds `check_span_order`, which errors in `transform_vsm_sections_to_body!`, `match_aero_sections_to_structure!` and `update_vsm_wing_from_structure!` on a wing whose sections run -y to +y, and removes the reversed-order handling from the live-polar chord frame (`spanwise_corners`, the `step` argument of `chord_blend_weights!`). With that check in place the per-panel span sign is +1 on every panel, and what is left of the opposite-order handling sits on the symbolic force path, which #308 does not touch:

- `panel_span_signs` (`src/aero_modes/common.jl`), and the `orient` it feeds into `VortexStepMethod.panel_axes` from `wagner_reference_frame`, `src/aero_modes/continuous.jl`, `src/aero_modes/pressure.jl` (`build_panel_force_eqs`), `src/kernel_backend/assembly.jl` and `test/test_continuous_modes.jl`.
- `AeroPanel(..., orient)` in `src/components/components.jl` and `panel_force_eqs`' `orient` argument.
- The `:_down` kernel key in `src/kernel_backend/assembly.jl` (`orient > 0 ? :_up : :_down`), which is dead once `orient` is always +1.
- `test/test_flow_curvature.jl`'s `orientation = [sign(dot(state.y_airf[i], panels[i].y_airf)) ...]`, always +1.

Stack on #308: it needs `check_span_order` to make `orient = 1` an invariant rather than an assumption. Removing `orient` changes the generated equations and kernel keys, so rebuild the model caches and run the kernel-backend and backend-parity tests.

1-Bort-1:
### TL;DR

`panel_span_signs` and the `orient` argument it fed into `VortexStepMethod.panel_axes` are gone from the symbolic force path, along with the `:_down` aero panel kernel. With #308's `check_span_order` refusing any wing whose sections run -y to +y, that sign is +1 on every panel, so the code handling the other order could never run.

### What changed

`panel_axes` has `orient=1` as its default, and VSM builds its own panel axes that way, so every call now leaves the argument off: `panel_force_eqs`, `build_panel_force_eqs`, `AeroPanel` and `wagner_reference_frame` all lose it. The kernel backend keys a wing's panels `:aero_panel_<wing>` instead of `_up`/`_down`, and the kernels already shared one compiled kernel per key, so a wing now builds exactly one. The removal also leaves `spanwise` unused in `add_aero_panels!` and `wagner_reference_frame`, so those two locals are gone too.

Why +1 is guaranteed and not just assumed: `panel_span_vector` points from section 2 to section 1, and `check_span_order` makes section 1 lie further along `spanwise_direction`. So `dot(span_vec, spanwise) >= 0`, which is the +1 branch of the deleted function.

Tests: `test_flow_curvature.jl` used to measure each panel's sign against VSM (`orientation = sign(dot(...))`) and multiply both the `y_airf` comparison and VSM's pitch rates by it. It now compares against VSM's `y_airf` and pitch rates directly, so a panel whose sign flipped would fail instead of being absorbed. `test/util.jl`'s `evaluate_panel_equations` and `test_continuous_modes.jl` just drop the argument.

Nothing user-visible changes: every symbol touched is private, and every wing `check_span_order` accepts gets the same equations as before. So there is no changelog entry.

### Where I would push back

`check_span_order` compares the first and last *unrefined* sections, while the deleted sign was taken per *refined* panel. A wing whose leading edge runs back on itself partway along the span, for example a tip curling past its neighbour, passes the check yet has a panel that used to get -1. Such a panel now gets VSM's own axes on that panel, which is what the numeric solver already used. I have not checked every fixture for such a panel; the two tests that compare panel axes against VSM directly (`test_continuous_modes.jl`, `test_flow_curvature.jl`) pass.

### Verification

- [x] Reproduced first: n/a — deletes code that `check_span_order` makes unreachable; nothing to reproduce
- [x] Monolith, on the rebased branch (juliaserver): `test_match_aero_sections.jl` 147/147, `test_continuous_modes.jl` 92/92, `test_flow_curvature.jl` 76/76, `test_unsteady_aero.jl` 42/42, `test_pressure_aero.jl` 100/100, `test_backend_parity.jl` 3/3, `test_analytic_jacobian.jl` 14/14
- [x] `KernelBackend`, same branch: `test_continuous_modes.jl` 92/92, `test_flow_curvature.jl` 36 pass + 10 skipped (its `has_panel_readout` skips, which the kernel always takes), `test_unsteady_aero.jl` 42/42, `test_pressure_aero.jl` 100/100, `test_flap_aero.jl` 35/35
- [x] Model caches rebuilt: each run used a fresh `mktempdir()` data path
- [ ] Local full suite (`agent ci-local`): started, not finished when this card was written · GitHub CI: running
- Docs build: not run locally. The only docs change removes the `panel_span_signs` `@docs` entry, and `grep` finds no remaining `@ref` to it. REUSE lint: not run locally (`reuse` is not installed on the box), and no files were added.
- Benchmark: n/a
- Risk: a wing that passes `check_span_order` but has one refined panel running against the span; see above.

### Scope

+32 / -69 across 9 files, all deletion of `orient`/`panel_span_signs` and the callers' argument lists. Stacked on #308 (`check_span_order`), which must merge first.

Closes #353 · task `SymbolicAWEModels.jl-353`


1-Bort-1 (commented):

**Verdict: APPROVE WITH COMMENTS** · 0 inline, 0 off the diff

**Good**

- Matches the card: `panel_span_signs`, the `orient` argument, the `_up`/`_down` kernel keys and the unused `spanwise` locals are all removed, and nothing else is changed; checked with `rg`, which finds no remaining `panel_span_signs`, `:_up` or `:_down` in src, test or docs.
- The `@docs` entry is removed along with the function, so `make.jl` cannot fail on a missing docstring (§2: remove what a removed feature used).
- The `spanwise` parameter is rightly kept in `panel_force_eqs`, which still passes it to `VortexStepMethod.panel_force_directions` (common.jl:362); the pressure and continuous callers and `AeroPanel` still need it.
- The `test_flow_curvature.jl` change makes the test stricter: the `orientation` factor that used to absorb a flipped panel is gone, so a sign mismatch against VSM's `y_airf` or pitch rates now fails the test.
- Scope is +32/-69, almost all deletion; no changelog entry is right because every touched symbol is private (none appear in the exported docs).
- `check_span_order` (vsm_refine.jl:103) runs on every setup path (vsm_refine.jl:98/243/472, common.jl:1473), so the +1 claim holds for the first-to-last span order it checks.

**Not good**

- The TL;DR says the other order 'could never run', but the pushback section admits it could: a refined panel that runs back against the span still passes the unrefined first/last check. State the invariant as what is actually enforced.
- For such a back-running panel this PR does change behaviour: it now gets VSM's axes instead of flipped ones. That is arguably a fix, since it now matches the numeric solver, but it is not 'nothing user-visible' for that geometry, and no fixture was checked for it.
- If per-panel +1 is meant to be the invariant now, `check_span_order` could check refined panels instead of the endpoints, or an issue could track it; the card only notes the gap.
- The comment kept at test_flow_curvature.jl:366-368 is three lines; §3 asks for at most one line.
- `agent ci-local` and GitHub CI had not finished when the card was written; the evidence is targeted juliaserver runs only.
- This is stacked on #308, so a `git diff main` shows 105 files; review it against #308's head, and merge #308 first.

1-Bart-1 (approved):

Issue #356, opened 2026-09-16
Wing ref points that are not wing frame members give a wrong body frame silently


1-Bort-1:
`setup_wing_frame!` (src/system_structure/system_structure_core.jl, RIGID branch) and `init_body_frame_from_ref_points!` build the body frame with `calc_particle_dynamics_wing_frame`, which reads `pos_w` of the `z_ref_points`/`y_ref_points`/`origin` points. Before that call only points with `wing_frame_member(point, wing.idx)` get `pos_w .= pos_cad`.

A ref point that is not a frame member (a `DYNAMIC` bridle point, or a point with no `wing=`, which is `wing_idx == 0` since 0.18) is read at whatever `pos_w` holds (zeros when it is fresh). `R_b_to_c` then comes out wrong, and so does `com_offset_b = R_b_to_c' * (com_cad - origin_cad)`, with no error or warning. The `origin` itself is read from `pos_cad`, so only the axes go wrong.

Found while bringing RamAirKites.jl onto 0.18 (OpenSourceAWE/RamAirKites.jl#45), where the wing frame refs have to name the `BODY_STATIC` wing points.

Suggested fix: when the refs resolve, check that every point they name is a `wing_frame_member` of that wing, and throw an `ArgumentError` naming the point otherwise. A test should build a wing whose `y_ref_points` names a `DYNAMIC` point and expect the error.

PR #357, opened 2026-09-16
Take a rigid wing's station moments and twist mass from its panels, and count its mass in total_mass

1-Bort-1:
### TL;DR

A rigid wing with fewer stations than VSM sections summed each station's twist moment over VortexStepMethod's left-edge section bins, which gives a symmetric wing asymmetric stations. A wing that carries its mass on its body left every station with zero twist inertia and `total_mass` at 0. Stations now take the panels nearest them, for the moment and for a share of the body mass, and `total_mass` counts a rigid wing's `mass`.

### What was wrong

Found bringing RamAirKites.jl onto 0.18 (OpenSourceAWE/RamAirKites.jl#45): a 41-section OBJ ram-air wing, 4 stations, its 0.9 kg mass on the wing body and none on its points.

**Moments.** VortexStepMethod 5 does not refine an OBJ wing, so `compute_refined_panel_mapping!` files panel i under section i, its left edge. `vsm_aero_coeffs` summed a station's twist moment over `moment_coeff_unrefined_dist` at the sections `compute_spatial_station_mapping!` gave it. The last section never has a panel, and the centre panel goes to whichever station owns the centre section: stations 1–4 got 10/11/10/9 panels. Forcing 10/10/10/10 in the running model (on released 0.18, point-mass wing) took `|acc_w|` at t = 2 s from 11.99 to 4.70 m/s² and let it run 10 s instead of failing at 6.11 s.

**Twist inertia.** `station_dynamics` uses ⅓·m·L² with m the sum of the station's point `extra_mass`. With the mass on the body that is 0, and `find_steady_state!` stops at t = 0 with NaN in the twist rows.

**total_mass.** It summed points only, so the same wing read 0.

### What changed

- `compute_spatial_station_mapping!` also gives every panel to the station nearest the centre of its corners (`Station.panel_idxs`). The linearized moment output sums `moment_coeff_dist` over those panels. Twist still goes to sections as before.
- The same function sets `Station.body_mass`: the wing mass that sits on no frame point, times the station's share of the panel area. Both backends add it to the points' mass in the twist inertia. For the ram-air wing that is ≈ 0.225 kg per station, the inertia its old 4 × 0.05625 kg points gave. A wing whose mass is on its points (every existing model) gets 0 and is unchanged.
- `total_mass` skips a rigid wing's frame points and adds the wing's `mass` once. The frame points carry no gravity of their own; their mass is already in `wing.mass`.
- `test_twist_alignment.jl` ramps the steering until a station twists past 0.1 rad, capped at 100 steps, and checks the chord alignment there. The old fixed 60-step ramp left the kernel backend at ±0.08 rad under the new panel sums, below the test's 0.1 rad bar. A longer or stronger ramp makes the monolith's VSM solve diverge, because the monolith already runs to its π/2 stop there. The bar and the tolerances are unchanged. Both backends now stop at 0.12–0.13 rad with the chord aligned to about 1e-13°.

### Verification

- [x] Reproduced first on released 0.18 with RamAirKites.jl's equilibrium procedure: `|acc_w|` 11.99 m/s² at 2 s and `MaxIters` at 6.11 s (point-mass wing); `Solver unstable at t=0.0` (mesh body)
- [x] With this branch and the mesh body: `|acc_w|` 4.27 m/s² at 2 s, aero force 276 N, runs the full 10 s
- [x] `test/test_multi_section_group.jl` 38/38 (panel partition, nearest-station ownership, left/right symmetry, `body_mass` share, `total_mass`); the `total_mass` assertions fail on main
- [x] `test_twist_alignment` on both backends after merging main (kernel stops at +0.12 rad, monolith at −0.13 rad)
- [x] `test_aero_modes` 809, `test_wing` 171, `test_backend_parity` 3, `test_static_twist` 11, `test_twist_alignment` 13 (juliaserver)
- Risk: a station on a wing with fewer refined panels than stations gets no panel and so no twist moment, where before the section it owned might have had one.

### Scope

Three commits, one per cause above, a merge of main, and the twist-alignment ramp. The neighbouring ref-point trap (#356) is not fixed here. OpenSourceAWE/RamAirKites.jl#45 needs this and waits for its release.

From OpenSourceAWE/RamAirKites.jl#45 · task `RamAirKites.jl-45`


1-Bart-1:
Merge in latest main and fix failing test @1-Bort-1

1-Bort-1:
Picked up as `SymbolicAWEModels.jl-357`, on the branch this pr already has.

1-Bort-1:
**Could not hand this over.** Error: cannot check out agent/RamAirKites.jl-45 for SymbolicAWEModels.jl-357: fatal: 'agent/RamAirKites.jl-45' is already used by worktree at '/home/agent/worktrees/RamAirKites.jl-45/.agent/upstream/OpenSourceAWE__SymbolicAWEModels.jl'

1-Bart-1 (approved):

